PR #36, opened 2022-11-23
Task: use 1ES compliant GitHub runner

### LINKED ISSUE #35: Add 1ES hosted runner
Please configure a 1ES hosted runner so that we can do cool things

MIchaelMainer:
Closes #35 

Let me know if you see any issues running the GitHub Actions especially if you find yourself waiting for a runner for too long.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -25,7 +25,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


gavinbarron:
@waldekmastykarz it is. We'll need to monitor these actions as we use them and get solid with these new runners. 
When they are hung it seems that cancelling and restarting the job is picking up the runner quickly

MIchaelMainer:
There is some brittleness in 1es runner system. If we encounter hung runners, cancel [all actions that are using the 1es runners ](https://github.com/microsoftgraph/msgraph-developer-proxy/actions) before resubmitting the job.

PR #65, opened 2022-12-06
Adds putting proxy files in a proxy-specific folder. Closes #40

### LINKED ISSUE #40: Put proxy-specific files in a proxy-specific folder in ~/.config
Right now, we put the rootCert.pfx file in the `~/.config` folder which is a generic folder that many apps on macOS use to store their config. By looking at the file, it's not clear that it belongs to the proxy, and could lead to folks deleting it. We should see if it's possible to adjust the config path to `~/.confg/msgraph-developer-proxy/` so that it's clear which files belong to the proxy and are safe to delete after uninstalling the proxy.

waldekmastykarz:
Adds putting proxy files in a proxy-specific folder. Closes #40

Because the config path isn't exposed outside of the ProxyServer, implemented custom certificate cache handler where we can specify the config path including an app-specific folder.

### DIFF
--- a/msgraph-developer-proxy/CertificateDiskCache.cs
+++ b/msgraph-developer-proxy/CertificateDiskCache.cs
@@ -1,0 +1,120 @@
+using System.Reflection;
+using System.Security.Cryptography.X509Certificates;
+using Titanium.Web.Proxy.Helpers;
+using Titanium.Web.Proxy.Network;
+
+namespace Microsoft.Graph.DeveloperProxy;
+
+// based on https://github.com/justcoding121/titanium-web-proxy/blob/9e71608d204e5b67085656dd6b355813929801e4/src/Titanium.Web.Proxy/Certificates/Cache/DefaultCertificateDiskCache.cs
+public sealed class CertificateDiskCache : ICertificateCache {
+    private const string DefaultCertificateDirectoryName = "crts";
+    private const string DefaultCertificateFileExtension = ".pfx";
+    private const string DefaultRootCertificateFileName = "rootCert" + DefaultCertificateFileExtension;
+    private string? rootCertificatePath;
+
+    public X509Certificate2? LoadRootCertificate(string pathOrName, string password, X509KeyStorageFlags storageFlags) {
+        var path = GetRootCertificatePath(pathOrName, false);
+        return LoadCertificate(path, password, storageFlags);
+    }
+
+    public void SaveRootCertificate(string pathOrName, string password, X509Certificate2 certificate) {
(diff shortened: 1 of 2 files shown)


garrytrinder (approved):
Looks good to me and I can see that you've already updated the [Uninstall](https://github.com/microsoftgraph/msgraph-developer-proxy/wiki/Uninstall) page in the wiki 👍

waldekmastykarz:
I have't updated the Wiki yet. That said, it's a good reminder that we should after we merge this PR

msgraph-developer-proxy/CertificateDiskCache.cs
```diff
@@ -0,0 +1,120 @@
...
+                rootCertificatePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), proxyConfigurationFolderName);
+            }
+            else if (RunTime.IsMac) {
+                rootCertificatePath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), proxyConfigurationFolderName);
+            }
+            else {
+                var assemblyLocation = GetType().Assembly.Location;
+
+                // dynamically loaded assemblies returns string.Empty location
+                if (assemblyLocation == string.Empty) assemblyLocation = Assembly.GetEntryAssembly().Location;
```
gavinbarron, line 98: Do we need to use a safe accessor here?

msgraph-developer-proxy/CertificateDiskCache.cs
```diff
@@ -0,0 +1,120 @@
...
+
+        var certPath = Path.Combine(path, DefaultCertificateDirectoryName);
+        if (create && !Directory.Exists(certPath)) Directory.CreateDirectory(certPath);
+
+        return certPath;
+    }
+
+    private string GetRootCertificateDirectory(bool create) {
+        if (rootCertificatePath == null) {
+            var proxyConfigurationFolderName = "msgraph-developer-proxy";
```
gavinbarron, line 83: If we're keeping other path fragments as const elements at the top of the file perhaps this belongs with them?

msgraph-developer-proxy/CertificateDiskCache.cs
waldekmastykarz: Good catch. Fixed

gavinbarron (approved):

garrytrinder:
I updated the [wiki page](https://github.com/microsoftgraph/msgraph-developer-proxy/wiki/Uninstall) with the new path.

PR #69, opened 2022-12-09
Adds contextual guidance for lack of $select

waldekmastykarz:
Adds contextual guidance for GET requests that don't use the `$select` parameter. Shows the guidance for requests that have been intercepted and passed thru.

image


gavinbarron (approved):
Nice!

Issue #54, opened 2022-12-02
Extend the proxy to support URLs other than Graph, eg. SharePoint REST/CSOM, MS Search, etc.


jansenbe:
Today when it comes to SharePoint/OneDrive applications we still see a many applications that take aa hybrid approach in which they use Graph for core scenarios but fall back to either SharePoint REST or SharePoint CSOM for the operations that covered via Graph APIs. 

In order to test such applications having support for SharePoint REST/CSOM would really help. The proxy could dynamically detect SharePoint URLs being hit (*.sharepoint.com, *.sharepoint.us,...) and that would work for the majority of cases. For vanity URL tenants ideally there's an option to specify the wildcard URL on which the proxy will take action.

This is a great tool to help developers, looking forward to see it's scope widen by supporting SharePoint REST/CSOM calls.

waldekmastykarz:
@gavinbarron that would require us to extend the list of domains, with SharePoint domains across the different clouds and allow using wildcards, correct?

gavinbarron:
@waldekmastykarz yes, it would require a bit of re-work on how we determine if a request is passed to the proxy behaviors or passed through. 

We should be able to use a regex or some other pattern match for this. 

As to extending the "cloud" being proxied, that should be reasonably straightforward.

waldekmastykarz:
Specifically to support SharePoint URLs, what if we'd use `.EndsWith()` rather than equals when we come across a URL with an asterisk? That should be faster than converting a string to a regex and matching it against every URL.

wobba:
There are other "internal" API REST URL's in M365 which would benefit from this proxy in terms of 1p dev `*.office.com` based ones.

waldekmastykarz:
Great suggestion @wobba and thank you for making us aware of other URLs for which the Proxy would be relevant

gavinbarron:
I think that based on the feedback from @wobba we should be looking at pattern matching for the domains to be intercepted rather than a `.EndsWith()`

waldekmastykarz:
[...]

PR #70, opened 2022-12-09
Adding multi-url-support specs

sebastienlevert:
Relates to #54 

Adds the specs for supporting multiple URLs in the proxy and the supporting configuration files required.

### DIFF
--- a/specs/multi-url-support.md
+++ b/specs/multi-url-support.md
@@ -1,0 +1,112 @@
+# Support for any URLs via the Microsoft Graph Developer Proxy
+
+Some applications are using an hybrid approach when it comes to consuming APIs. Some APIs are available on Microsoft Graph while some are not. To enable developers to fully test their scenarios with the Microsoft Graph Developer Proxy, the proxy should support acting as a reverse proxy for more than just one URL at a time, and include URLs that are outside of the Microsoft Graph. This can also be used by internal teams to test their APIs on internal endpoints.
+
+## History
+
+| Version | Date | Comments | Author |
+| ------- | -------- | ----- | --- |
+| 1.0 | 2022-12-09 | Initial specifications | @sebastienlevert
+
+
+## Configuration file
+
+A configuration file should be available for developers who want to support scenarios outside of Microsoft Graph. The file should be at the root of the proxy working folder and named `config.json`. This file should have default RegEx values for all supported Microsoft Graph clouds and SharePoint URLs.
+
+This configuration can also be extended by "any" URLs to support 1P endpoints, vanity URLs, custom services, etc.
+
+| Property | Type | Comments |
+| ------- | -------- | ----- |
+| id | String (unique) | This can be any string value but needs to be unique |
+| regEx | String | String value of the RegEx to use |
+| isWildcard | Boolean | If the value is `true`, the RegEx gets an appended wildcard token. If `false`, only the RegEx value should be considered |
+
+> **Note**
+> These RegEx need to be validated in the context of the App
+
+```json
+{
+  "endpoints": [
+    { 
+      "id": "microsoftGraph",
+      "regEx": "https:\/\/(canary.)?((graph|dod-graph).microsoft.(com|us)|microsoftgraph.chinacloudapi.cn)",
+      "isWildcard": true
+    },
+    {
+      "id": "sharePoint",
+      "regEx": "https:\/\/(.*).sharepoint(-df)?.(com|us)",
+      "isWildcard": true
+    },
+    {
+      "id": "customService",
+      "regEx": "https:\/\/customService.azurewebsites.net",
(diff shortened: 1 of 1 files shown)


wobba:
Love it! 🙂

specs/multi-url-support.md
```diff
@@ -0,0 +1,112 @@
...
+| ------- | -------- | ----- |
+| id | String (unique) | This can be any string value but needs to be unique |
+| regEx | String | String value of the RegEx to use |
+| isWildcard | Boolean | If the value is `true`, the RegEx gets an appended wildcard token. If `false`, only the RegEx value should be considered |
+
+> **Note**
+> These RegEx need to be validated in the context of the App
+
+```json
+{
```
waldekmastykarz, line 28: I suggest to make it easier and use the following structure to define the list of hosts to watch:

```json
{
  "hostsToWatch": [
    "graph.microsoft.com",
    "graph.microsoft.us",
    ...
    "*.sharepoint.com"
  ]
}
```

Rather than using regex, we'd use the existing wildcard-based pattern matching, that we have already implemented in matching URLs for mock responses, allowing us to reuse existing code, and implement this quicker. With the existing pattern matching, we check for the presence of `*` in a string. If we find it, we use regex matching replacing `*` with `.*` (basically any character). If we don't, we use a string equality comparison to determine the match.

Requiring users to build regex makes it harder to use the proxy. Writing regex is hard. Especially, because there's no easy way for folks to debug the queries, it will force them to use other tools, further adding complexity.
Writing good regex is hard, and this is important because we'll be using these patterns to match against every request we intercept and then again to check if the response should be mocked. So if the regex is written badly, it might significantly slow down the proxy.

Using a simple list of hosts instead of regex and complex objects has another benefit. When we add support for importing mocks from a HAR file, we could automatically detect host names that we don't have in the list and add them to ensure that these requests are properly captured as well. If we work with regex and objects , there's a good chance we'll end up with duplicate definitions which could lead to undesired side-effects and hard to debug issues.

specs/multi-url-support.md
```diff
@@ -0,0 +1,112 @@
...
+    }
+  ]
+    
+  }
+}
+```
+
+## Remove the `--cloud` option on the Proxy
+
+This new capability renders the `--cloud` option obsolete on the Proxy. We should deprecate the capability, remove associated documentation and once the Proxy hits its next major version, introduce the breaking change. Until then, if the option is used, notify the developer that this option now is obsolete and will use the `config.json` file as its source of truth for URLs it should proxy.
```
waldekmastykarz, line 58: Since we're in a preview stage, I suggest we simply remove the option and the corresponding documentation, without waiting for a major version.

specs/multi-url-support.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Mocking responses
+
+Responses can be mocked and should be updated to include either the absolute URL for the response to be returned or a RegEx. This RegEx would be used in scenarios where a developer want to test the same set of responses on multiple clouds or on different environments. Is is possible to use the `id` property of the endpoint to reuse the base RegEx with using `{valueOfTheId}` and append paths.
```
waldekmastykarz, line 62: Following the simplified implementation of URLs to intercept, I suggest we use the existing wildcard mechanism, and if users want to provide a single mock for multiple Graph API URLs, they use for example `https://*graph*/...` which should capture all Graph URLs across the different clouds. Depending on the need, we can provide recommended examples in the documentation.

waldekmastykarz (changes requested):

specs/multi-url-support.md
gavinbarron: I like this reasoning to keep it simple. You're absolutely right about RegEx, it's easy to get it wrong because it's arcane and varies depending on the run time. I really like the idea of supporting simple wildcard matching.

specs/multi-url-support.md
gavinbarron: Agreed. This is the point in time where we can and should make breaking changes like this until we reach a solid GA point.

specs/multi-url-support.md
```diff
@@ -0,0 +1,112 @@
...
+## History
+
+| Version | Date | Comments | Author |
+| ------- | -------- | ----- | --- |
+| 1.0 | 2022-12-09 | Initial specifications | @sebastienlevert
+
+
+## Configuration file
+
+A configuration file should be available for developers who want to support scenarios outside of Microsoft Graph. The file should be at the root of the proxy working folder and named `config.json`. This file should have default RegEx values for all supported Microsoft Graph clouds and SharePoint URLs.
```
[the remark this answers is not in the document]
sebastienlevert: Agree!

waldekmastykarz (approved):
Let's ship it!

PR #87, opened 2022-12-16
Extends Proxy to support other URLs. Closes #54

### LINKED ISSUE #54: Extend the proxy to support URLs other than Graph, eg. SharePoint REST/CSOM, MS Search, etc.

waldekmastykarz:


msgraph-developer-proxy/responses.json
```diff
@@ -1,7 +1,7 @@
 {
   "responses": [
     {
-      "url": "/v1.0/me",
+      "url": "https://graph.microsoft.com/v1.0/me",
```
gavinbarron, line 4: Should we have an example url in the responses that shows using a wildcard in the hostname? 

e.g. "url": "https://graph.microsoft.*/v1.0/me",

gavinbarron (commented):
Implementation looks good, just one question about our sample responses for mocking.

msgraph-developer-proxy/responses.json
waldekmastykarz: We've got `https://graph.microsoft.com/v1.0/users/*/photo/$value` and `https://graph.microsoft.com/v1.0/users/*`. Do you think it would be valuable to add one with a wildcard in the hostname as well? Btw, we're not shipping this file in the release, so it's only in this repo for testing. Otherwise, it's something that we'd explain more in detail in our docs.

gavinbarron: You're right, I'd forgotten that we were not shipping this file, this is all good IMO.

gavinbarron (approved):

PR #89, opened 2022-12-19
Adds proxy header. Closes #83

### LINKED ISSUE #83: Add Proxy headers to requests we pass thru to Graph
Using the Via header to track Proxy usage.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Via

```
Via: 1.1 graph-proxy/0.3.0
Via: <protocol-version> <pseudonym>
```

waldekmastykarz:
Adds proxy header. Closes #83

### DIFF
--- a/msgraph-developer-proxy/ChaosEngine.cs
+++ b/msgraph-developer-proxy/ChaosEngine.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Diagnostics;
 using System.Net;
+using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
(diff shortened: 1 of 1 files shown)


msgraph-developer-proxy/ChaosEngine.cs
```diff
@@ -269,6 +288,12 @@ public class ChaosEngine {
                 UpdateProxyResponse(e, responseComponents, matchingResponse);
         }
 
+        private static void AddProxyHeader(Request r) {
+            if (r.Headers is not null) {
```
gavinbarron, line 292: Should we initialize the headers collection if it is null? 
For graph requests we know that will be a 401, so should we even bother to forward the request to graph?

gavinbarron (approved):

PR #92, opened 2022-12-20
docs: initial proposal for plug-in system

gavinbarron:
This spec provides a proposal to cover:
- Loading plugins
  - Includes a proposal for implementation
- Providing plugin specific configuration via the appsettings.json file
- Providing plugin specific command line options


specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+
+## History
+
+| Version | Date | Comments | Author |
+| ------- | -------- | ----- | --- |
+| 1.0 | 2022-12-19 | Initial specifications | @gavinbarron |
+
+## Configuration file
+
+The existing `appsettings.json` file will be used to provide configuration. Both for determining which request handlers are loaded into the proxy and providing configuration for each of the loaded request handlers. The set of request handlers to be loaded will be specified as an array of objects at the root of the configuration file in the `handlers` property. Each handler must supply a `configSection`, `handerPath`, `name`, and `urlsToWatch` in the handlers array.
```
waldekmastykarz, line 14: What if we automatically loaded plugins from all dlls in the proxy folder? Folks would only need to specify them in the config file, if the plugin requires config. When they don't need a config, they can just drop the dll and off they go.

waldekmastykarz, line 14: Shall we name the collection `plugins` rather than `handlers` just so that we cater to a broader set of scenarios, beyond what we're envisioning now? Also, aren't handlers referring just to the code that handles an intercepted request, whereas we also wanted to support behaviors, like rate limiting?

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+- `configSection` is a string for which there must be a matching object property at the root of the configuration file. If a handler does not provide any configuration options then an empty object must be supplied.
```
waldekmastykarz, line 15: Why do we require specifying an empty object if the plugin doesn't need any configuration? Isn't it making it harder than necessary for the users to bring in plugins?

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+- `handlerPath` a string which is the relative path to the assembly containing the class definition for the handler which will be loaded using reflection.
+- `name` a string identifier for the handler class, allows developer to ship multiple handlers in a single assembly.
+- `urlsToWatch` is an array of strings defining the urls for which this hander will watch, this behavior is defined in the [Multi URL support spec](./multi-url-support.md).
```
waldekmastykarz, line 18: Won't this lead to repeating the list of URLs to watch for each and every plugin? Why not have a proxy-wide list of URLs to watch?

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Use this method to add any command line options to the proxy
+     */
+    public abstract void AddCommandLineOptions(RootCommand command);
+
+    /**
+     * Read options from command line args and update internal configuration
+     */
+    public abstract int UpdateConfiguration(InvocationContext context);
```
waldekmastykarz, line 72: What is the `int` value that we're returning here?

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+Handlers can set the response which will be provided by the proxy using the `GenericResponse` method of the provided `SessionEventArgs` object, when a handler sets the response it should also set the `IsComplete` property of the provided `ResponseState` to prevent subsequently executed handler from modifying the response.
+
+```cs
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.Text.RegularExpressions;
+using Microsoft.Extensions.Configuration;
+using Titanium.Web.Proxy.EventArguments;
+
+public abstract class ProxyHandler
```
waldekmastykarz, line 51: How would a plugin following this setup allow us to bring it additional functionality like rate limiting behaviors?

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+| Member | kind | purpose | Notes |
+| -------| ---- | ------- | ------|
+| Name | Read only property | Provides a name to allow developers to ship multiple handlers in a single assembly | Should match the class name and be unique per assembly |
+| AddCommandLineOptions | method | Provides typed commandline inputs and validation | Use `Option<T>` from `System.CommandLine` to define command line options |
+| UpdateConfiguration | method | Overwrite configuration defaults with supplied command line options | Read values for supplied options from InvocationContext |
+| ExecuteInternal | method | Handler specific logic to be applied to matching requests | `ResponseState.HasBeenSet` should be set to true if the response has been set to prevent unwanted changes to the response |
+
+### Plug-in system
+
+The Hander plug-in system will leverage reflection to load the custom logic at runtime via the supplied configuration.
```
waldekmastykarz, line 113: `handler` (typo)

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+     * better handled by configuring the handler to run earlier in the set of configured handlers
+     */
+    protected virtual bool ShouldTrigger(SessionEventArgs e, ResponseState state) =>
+        !state.HasBeenSet &&
+        this._urlsToWatch.Any(u => u.IsMatch(e.HttpClient.Request.RequestUriString));
+
+    /**
+     * Public entrypoint called by the Proxy for each request.
+     */
+    public void Excecute(SessionEventArgs e, ResponseState state)
```
waldekmastykarz, line 93: Here, we should also pass an instance of the logger, so that a plugin can write logging information, console or whatever output is configured in the Proxy. By specifying a list of types of logging messages, we'll also ensure consistent logging across plugins.

specs/request-handler-plugins.md
```diff
@@ -0,0 +1,116 @@
...
+## Configuration file
+
+The existing `appsettings.json` file will be used to provide configuration. Both for determining which request handlers are loaded into the proxy and providing configuration for each of the loaded request handlers. The set of request handlers to be loaded will be specified as an array of objects at the root of the configuration file in the `handlers` property. Each handler must supply a `configSection`, `handerPath`, `name`, and `urlsToWatch` in the handlers array.
+- `configSection` is a string for which there must be a matching object property at the root of the configuration file. If a handler does not provide any configuration options then an empty object must be supplied.
+- `handlerPath` a string which is the relative path to the assembly containing the class definition for the handler which will be loaded using reflection.
+- `name` a string identifier for the handler class, allows developer to ship multiple handlers in a single assembly.
+- `urlsToWatch` is an array of strings defining the urls for which this hander will watch, this behavior is defined in the [Multi URL support spec](./multi-url-support.md).
+
+```json
+  "handlers": [
```
waldekmastykarz, line 21: Should we add the ability to temporarily disable a plugin without having to remove the whole config block from the config? That would be useful in cases when you want to disable a specific plugin/guidance when testing a specific app.

specs/request-handler-plugins.md
waldekmastykarz, line 93: Shall we also allow plugins to process the response of a request that we passed through to Graph? Right now, we have the limitation that we [stop our processing pipe](https://github.com/microsoftgraph/msgraph-developer-proxy/blob/e8ca93943a8ab101cff7dbc1d9ad50c7cd272345/msgraph-developer-proxy/ChaosEngine.cs#L311-L315) when we decided that the request should be passed to Graph. If that request got throttled, our SDK guidance wouldn't run. If we're to allow for this, we'd need to consider using an event-based handling rather than a single method.

waldekmastykarz (commented):
Should we consider implementation based on events and event handlers? In the proxy we'd emit a number of events, to which plugins could subscribe if they choose to handle them. That way, we'd have more flexibility in the future: if we wanted to add a new extensibility point, we'd introduce a new event rather than having to rewrite the whole plugin system with additional abstract or virtual members, that folks might have to override even if they don't need it, and which would break existing plugins if not implemented.

waldekmastykarz:
One more thing that popped into my mind, shall we consider extending the scope of the spec to exposing shared utils that plugins can use, like `IsGraphRequest` so that we have consistency across the board and can avoid each plugin defining their own logic for it?

specs/request-handler-plugins.md
gavinbarron: I looked at that idea, it's certainly an option, the biggest risk/drawback is that if there are two plugins with conflicting dependencies it won't work. Imagine one plugin that uses Newtonsoft.Json @ v8 and a second that uses Newtonsoft.Json @ v13. Loading the assemblies directly out of the same folder would cause the app to crash when it failed to resolve one of the version of Newtonsoft.Json

gavinbarron: I like the name shift. 
Rate limiting is still going to need to intercept the requests to apply the metering, and potentially apply a response if the meter is exceeded.

specs/request-handler-plugins.md
gavinbarron: Totally valid, it makes implementation a little more complex in that we need to deal with a null config section, but that's simple enough.

specs/request-handler-plugins.md
gavinbarron: Yes, it does lead to repetition.
It also allows for plugins to provide URL specific behaviors. For example, with the plugins specifying the URLs that they will be acting on requests for we could have a set up where there is a Rate Limiting plugin for SharePoint URLs providing appropriate behavior for that domain and a Rate Limiting plugin for Teams APIs that provides Teams style rate limiting behavior running simultaneously.
Yes, it's a bit more cumbersome but ultimately far more powerful.

specs/request-handler-plugins.md
gavinbarron: Agreed! I was thinking about that use case yesterday when I was reading your output formatting options but failed to translate it into here.

specs/request-handler-plugins.md
gavinbarron: That's a great idea!

specs/request-handler-plugins.md
gavinbarron: In principle I like the idea of allowing a plugin to tap into the response being sent, be it from the underlying API or as a result of intercepting the request and setting a response in a plugin.

I'm just not sure what the use case for this would be.

specs/request-handler-plugins.md
gavinbarron: A return code for the proxy, non-zero on error to trigger the proxy to exit.

gavinbarron:
After our conversation this morning I'm very much on board with the idea of making the plugins event based and will work on updating this spec. 
Also, a big yes to having common utils that can be re-used by the plugins.

specs/request-handler-plugins.md
gavinbarron: This is irrelevant now, under the event-based model implementers would need to throw unhandled exceptions to trigger the proxy to exit.

specs/request-handler-plugins.md
gavinbarron: resolving under the move to the event-based model and order should cover the needs here

specs/request-handler-plugins.md
gavinbarron: I've incorporated a logger into the base event args we pass

gavinbarron: Resolved via change to event-based model

specs/request-handler-plugins.md
waldekmastykarz: Good point. Also, we'd have no way to determine their order. Let's stick with your proposal.

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+
+> It is strongly recommended that plugin implementers provide a default configuration in their code.
+
+Plugin classes may register a handler for any events provided in the `IPluginEvents` interface. Plugins registering a handler for the `Init` event should also register a handler for the `OptionsLoaded` event to consume the supplied options.
+
+All of the event arg types will inherit from a common base type which will supply common logic and utilities. A proposed implementation of the `PluginArgsBase` is provided here:
+
+```cs
+public abstract class PluginArgsBase {
+    protected PluginArgsBase(IConsoleLogger consoleLogger) {
```
waldekmastykarz, line 79: Shall we extend this parameter to `IProxyContext context` (so `public Plugin(IProxyContext context)`), which has a property named `Logger`? That way, we can have a place to pass more information in the future if need be, without breaking existing plugins. Also, for the logger interface, I suggest we keep the name generic, because the implementation decides where the output goes to.

waldekmastykarz, line 79: Rather than passing the logger/context along with each event, shall we move it to the `IProxyPlugin.Register` method, where the plugin could grab the reference and store for future use? It would simplify the EventArgs object that we need to build and pass on each event.

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+        ConsoleLogger = consoleLogger ?? throw new ArgumentNullException(nameof(consoleLogger));
+    }
+    public IConsoleLogger ConsoleLogger { get; set; }
+
+    public static bool IsSdkRequest(Request request) {
```
waldekmastykarz, line 84: Since these methods are static, shall we move them to a separate ProxyUtils class? That way, if we build a plugin that's not related to Graph and which doesn't need these methods, they're not a part of the implementation.

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+The `Register` method of a `IProxyPlugin` will be called after an instance is created and will provide an `ISet<Regex>` to define the urls the plugin wll watch, either using the default or a plugin specific set based on the supplied configuration and an `IConfigurationSection` if one was loaded.
+
+> It is strongly recommended that plugin implementers provide a default configuration in their code.
+
+Plugin classes may register a handler for any events provided in the `IPluginEvents` interface. Plugins registering a handler for the `Init` event should also register a handler for the `OptionsLoaded` event to consume the supplied options.
+
+All of the event arg types will inherit from a common base type which will supply common logic and utilities. A proposed implementation of the `PluginArgsBase` is provided here:
+
+```cs
+public abstract class PluginArgsBase {
```
waldekmastykarz, line 78: Shouldn't a base class for event payload inherit from EventArgs?

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+        return "https://aka.ms/move-to-graph-js-sdk";
+    }
+
+    public static string BuildApiErrorMessage(Request r) => $"Some error was generated by the proxy. {(IsGraphRequest(r) ? (IsSdkRequest(r) ? "" : BuildUseSdkMessage(r)) : "")}";
+
+    public static string BuildUseSdkMessage(Request r) => $"To handle API errors more easily, use the Graph SDK. More info at {GetMoveToSdkUrl(r)}";
+}
+```
+
+The details of the IConsoleLogger interface and implementation are out of the scope of this spec, except to say that they will provide plugins with a standardized mechanism for logging messages using preset formatting and categorization.
```
waldekmastykarz, line 102: Good point! For reference: we'd implement something like https://github.com/waldekmastykarz/console-output/blob/135c9751ffedbe1737d20f7f55910b0d50fa7439/Program.cs#L51 and then call it passing the message to log and its type. In the method, we should also allow passing multiple strings to allow printing multi-line messages.

waldekmastykarz (changes requested):
Awesome to see it come together! For completeness, let's also include a sample plugin implementation, so that we can more easily see how it would be used in practice, and if we properly separate concerns.

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+Plugin classes may register a handler for any events provided in the `IPluginEvents` interface. Plugins registering a handler for the `Init` event should also register a handler for the `OptionsLoaded` event to consume the supplied options.
+
+All of the event arg types will inherit from a common base type which will supply common logic and utilities. A proposed implementation of the `PluginArgsBase` is provided here:
+
+```cs
+public abstract class PluginArgsBase {
+    protected PluginArgsBase(IConsoleLogger consoleLogger) {
+        ConsoleLogger = consoleLogger ?? throw new ArgumentNullException(nameof(consoleLogger));
+    }
+    public IConsoleLogger ConsoleLogger { get; set; }
```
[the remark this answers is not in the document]
gavinbarron: It's part of the object passed to the plugin when the event is fired, not part of the plugin itself so needs to be public AFAIK.

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+Plugin classes must implement the `IProxyPlugin` interface and provide a parameterless constructor.
+
+```cs
+public interface IProxyPlugin {
+    string Name { get; }
+    void Register(IPluginEvents pluginEvents, IConfigurationSection? configSection);
+}
+
+public interface IPluginEvents {
+    event EventHandler<InitArgs> Init;
```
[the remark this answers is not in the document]
gavinbarron: I believe it will https://learn.microsoft.com/en-us/dotnet/csharp/event-pattern#event-delegate-signatures 
The key part being:
> Events are based on delegates and are multicast delegates. That supports multiple subscribers for any event source

Plugin implementers should be able to use `pluginEvents.Init += OnInit` to register their event sinks

specs/plugins.md
gavinbarron: That's not necessary anymore: https://learn.microsoft.com/en-us/dotnet/csharp/modern-events

specs/plugins.md
gavinbarron: That was a conclusion I reached as I worked on a sample implementation too!

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+
+    public static bool IsGraphRequest(Request request) {
+        return request.RequestUri.Host.Contains("graph", StringComparison.OrdinalIgnoreCase);
+    }
+    public static string GetMoveToSdkUrl(Request request) {
+        // TODO: return language-specific guidance links based on the language detected from the User-Agent
+        return "https://aka.ms/move-to-graph-js-sdk";
+    }
+
+    public static string BuildApiErrorMessage(Request r) => $"Some error was generated by the proxy. {(IsGraphRequest(r) ? (IsSdkRequest(r) ? "" : BuildUseSdkMessage(r)) : "")}";
```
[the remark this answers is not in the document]
gavinbarron: You're absolutely right.

gavinbarron:
Thanks for the review @waldekmastykarz this feels like it's coming together nicely.
I'm going to work on a sample implementation, likely using the `$select` guidance as a nice lightweight plugin

specs/plugins.md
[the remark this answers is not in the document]
gavinbarron: Confirmed as working via a spike implementation

specs/plugins.md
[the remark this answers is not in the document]
gavinbarron: This approach is now defunct as the `IProxyContext` is passed during registration

specs/plugins.md
```diff
@@ -0,0 +1,134 @@
...
+- `name` - required, a string identifier for the plugin class, allows developer to ship multiple plugins in a single assembly.
+- `configSection` - optional, a string for which there must be a matching object property at the root of the configuration file. If a plugin supplies a `configSection` and no corresponding property exists in the configuration file an error will be thrown during start up of the proxy.
+- `urlsToWatch` - optional, an array of strings defining the urls for which this plugin will watch instead of the `urlsToWatch` at the root of the config file, this behavior is defined in the [Multi URL support spec](./multi-url-support.md).
+
+```json
+  "plugins": [
+    {
+        "configSection": "randomErrors",
+        "pluginPath": "RandomErrorPlugin\\msgraph-proxy-handler.dll",
+        "name": "RandomErrorPlugin",
```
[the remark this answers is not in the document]
waldekmastykarz: Ah, got it 👍

specs/plugins.md
```diff
@@ -0,0 +1,173 @@
+# Support a "plug-in" model to provide extensibility
+
+To allow developers to provide unique handling behaviors for different APIs the Microsoft Graph will have a plug-in model provide an extension point for developer. This will provide separation of concerns between the software components intercepting network request and those providing behaviors.
```
waldekmastykarz, line 3:
```suggestion
To allow developers to provide unique handling behaviors for different APIs the Microsoft Graph Developer Proxy will have a plug-in model provide an extension point for developer. This will provide separation of concerns between the software components intercepting network request and those providing behaviors.
```

specs/plugins.md
```diff
@@ -0,0 +1,173 @@
...
+        "disabled": false
+    }
+  ],
+  "randomErrors": {
+    "rate": 50,
+    "allowedErrors": [ 429, 500, 502, 503, 504, 507 ]
+  }
+```
+
+Plugins will executed in the order that they are listed in the plugins configuration array. At least one plugin must be present in the configuration, otherwise the proxy will exit with an error informing the user that their configuration is invalid and contains no plugins.
```
waldekmastykarz, line 49:
```suggestion
Plugins will be executed in the order that they are listed in the plugins configuration array. At least one plugin must be present in the configuration, otherwise the proxy will exit with an error informing the user that their configuration is invalid and contains no plugins.
```

specs/plugins.md
```diff
@@ -0,0 +1,173 @@
...
+
+public interface IPluginEvents {
+    event EventHandler<InitArgs> Init;
+    event EventHandler<OptionsLoadedArgs> OptionsLoaded;
+    event EventHandler<ProxyRequestArgs> Request;
+    event EventHandler<ProxyResponseArgs> Response;
+}
+
+public interface IPluginContext {
+    ILogger logger { get; }
```
waldekmastykarz, line 72: Should it be `Logger` (uppercase `L`) as it's a public member?

waldekmastykarz (approved):
Awesome! Ship it! 🚀

specs/plugins.md
waldekmastykarz:
```suggestion
    ILogger Logger { get; }
```

PR #99, opened 2023-01-04
feat: plug in system implementation 

gavinbarron:
refactors existing proxy features into plugins:
- RandomErrorPlugin : Provides random failures
- MockResponsePlugin: provide pre-configured responses
- SelectGuidancePlugin: detects missing $select on GET requests
- SdkGuidancePlugin: detects raw REST calls without sdk-version headers 

Adds a basic console logger to decouple console writing from plugin code allowing for richer formatting in future with minimal plugin changes
Adds an abstractions library to provide shared code and interface definitions for communication between the proxy and plugins
Adds common utilities to ProxyUtils class
Renames a few classes to better align to their usage


msgraph-developer-proxy-abstractions/PluginEvents.cs
```diff
@@ -0,0 +1,89 @@
...
+    event EventHandler<ProxyResponseArgs> Response;
+}
+
+public class PluginEvents : IPluginEvents {
+    public event EventHandler<InitArgs>? Init;
+    public event EventHandler<OptionsLoadedArgs>? OptionsLoaded;
+    public event EventHandler<ProxyRequestArgs>? Request;
+    public event EventHandler<ProxyResponseArgs>? Response;
+
+    public void FireInit(InitArgs args) {
```
waldekmastykarz, line 74: Shall we follow the Microsoft naming convention here?

image

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/event

msgraph-developer-proxy/ProxyEngine.cs
```diff
@@ -0,0 +1,257 @@
...
+using Titanium.Web.Proxy;
+using Titanium.Web.Proxy.EventArguments;
+using Titanium.Web.Proxy.Helpers;
+using Titanium.Web.Proxy.Http;
+using Titanium.Web.Proxy.Models;
+
+namespace Microsoft.Graph.DeveloperProxy;
+
+public class ProxyEngine {
+
```
waldekmastykarz, line 18: Let's remove the unnecessary empty line

msgraph-developer-proxy/ProxyEngine.cs
```diff
@@ -0,0 +1,257 @@
...
+    private readonly ProxyConfiguration _config;
+    private ProxyServer? _proxyServer;
+    private ExplicitProxyEndPoint? _explicitEndPoint;
+    // lists of URLs to watch, used for intercepting requests
+    private ISet<Regex> _urlsToWatch = new HashSet<Regex>();
+    // lists of hosts to watch extracted from urlsToWatch,
+    // used for deciding which URLs to decrypt for further inspection
+    private ISet<Regex> _hostsToWatch = new HashSet<Regex>();
+
+
```
waldekmastykarz, line 30: Let's remove the unnecessary empty line

msgraph-developer-proxy/appsettings.json
```diff
@@ -1,4 +1,36 @@
...
+      "pluginPath": "GraphProxyPlugins\\msgraph-proxy-plugins.dll",
+      "name": "MockResponsePlugin",
+      "disabled": false
+    },
+    {
+      "configSection": "randomErrorsPlugin",
+      "pluginPath": "GraphProxyPlugins\\msgraph-proxy-plugins.dll",
+      "name": "RandomErrorPlugin",
+      "urlsToWatch": [
+        "https://graph.microsoft.com/v1.0/*",
```
waldekmastykarz, line 22: Shouldn't we allow testing all URLs with random errors?

msgraph-developer-proxy/appsettings.json
```diff
@@ -1,4 +1,36 @@
 {
+  "handlers": [
+    {
+      "pluginPath": "GraphProxyPlugins\\msgraph-proxy-plugins.dll",
+      "name": "SelectGuidancePlugin"
+    },
+    {
+      "pluginPath": "GraphProxyPlugins\\msgraph-proxy-plugins.dll",
+      "name": "SdkGuidancePlugin"
```
waldekmastykarz, line 9: We could trim this plugin to run only on Graph requests

msgraph-developer-proxy/appsettings.json
```diff
@@ -12,5 +44,12 @@
     "https://*.sharepoint.*/*_vti_bin/*",
     "https://*.sharepoint-df.*/*_api/*",
     "https://*.sharepoint-df.*/*_vti_bin/*"
-  ]
+  ],
+  "mocksPlugin": {
+    "mocksFile":  "mocks.json"
+  },
+  "randomErrorsPlugin": {
+    "rate": 99,
```
waldekmastykarz, line 52: Since this is a default value for all users, shall we leave it at the default 50 which we have right now?

msgraph-proxy-plugins/Guidance/SdkGuidancePlugin.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        _urlsToWatch = urlsToWatch;
+        _logger = context.Logger;
+
+        pluginEvents.Request += OnRequest;
+    }
+
+    private void OnRequest(object? sender, ProxyRequestArgs e) {
+        Request request = e.Session.HttpClient.Request;
+        if (_urlsToWatch is not null && e.ShouldExecute(_urlsToWatch) && WarnNoSdk(request))
```
waldekmastykarz, line 40: I think the semantics of this plugin are different comparing to what we have right now. Right now, we only show the message when the request was issued without the SDK and the request failed: https://github.com/microsoftgraph/msgraph-developer-proxy/blob/33e42621b4bc15c53669fd3e1f3be887053a9793/msgraph-developer-proxy/ChaosEngine.cs#L317-L323

To keep this behavior intact, we should subscribe to the response event where we can check if the response is different than 200. If the requests succeeds, it would be odd for us to advise users to use the SDK to handle errors, while there aren't any, right? 😄

msgraph-proxy-plugins/MocksResponses/MockResponse.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
```
waldekmastykarz, line 1: Should the folder name be `MockResponses` rather than `MocksResponses`?

msgraph-proxy-plugins/RandomErrors/RandomErrorPlugin.cs
```diff
@@ -0,0 +1,256 @@
...
+        session.GenericResponse(body ?? string.Empty, errorStatus, headers);
+    }
+
+    private string BuildThrottleKey(Request r) => $"{r.Method}-{r.Url}";
+
+    public void Register(IPluginEvents pluginEvents,
+                         IProxyContext context,
+                         ISet<Regex> urlsToWatch,
+                         IConfigurationSection? configSection = null) {
+
```
waldekmastykarz, line 177: Let's remove the unnecessary empty line

msgraph-proxy-plugins/MocksResponses/MockResponsePlugin.cs
```diff
@@ -0,0 +1,161 @@
...
+    private void OnOptionsLoaded(object? sender, OptionsLoadedArgs e) {
+        InvocationContext context = e.Context;
+        // allow disabling of mocks as a command line option
+        _configuration.NoMocks = context.ParseResult.GetValueForOption(_noMocks);
+    }
+
+    private void OnRequest(object? sender, ProxyRequestArgs e) {
+        Request request = e.Session.HttpClient.Request;
+        ResponseState state = e.ResponseState;
+        if (!_configuration.NoMocks && _urlsToWatch is not null && e.ShouldExecute(_urlsToWatch)) {
```
waldekmastykarz, line 81: Shouldn't we also check here if the response has already been set to avoid overwriting another plugin?

waldekmastykarz (changes requested):
So cool to see this taking shape! Let's do a few adjustments before we proceed.

msgraph-developer-proxy/PluginLoader.cs
```diff
@@ -0,0 +1,115 @@
...
+internal class PluginLoader {
+
+    public PluginLoader(ILogger logger) {
+        _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+    }
+
+    public PluginLoaderResult LoadPlugins(IPluginEvents pluginEvents, IProxyContext proxyContext) {
+
+        List<IProxyPlugin> plugins = new();
+        PluginConfig config = HandlerConfig;
```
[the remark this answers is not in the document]
gavinbarron: I'd been working on cleaning those up and missed that one.

msgraph-developer-proxy-abstractions/ProxyUtils.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using Titanium.Web.Proxy.Http;
+using Titanium.Web.Proxy.Models;
+
+namespace Microsoft.Graph.DeveloperProxy.Abstractions;
+public static class ProxyUtils {
+
+    public static string BuildApiErrorMessage(Request r) => $"Some error was generated by the proxy. {(IsGraphRequest(r) ? IsSdkRequest(r) ? "" : BuildUseSdkMessage(r) : "")}";
+
+    public static string BuildUseSdkMessage(Request r) => $"To handle API errors more easily, use the Graph SDK. More info at {GetMoveToSdkUrl(r)}";
```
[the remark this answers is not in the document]
gavinbarron: Funnily enough this one is actually emitted by the RandomErrorsPlugin in the body of the generated error response

msgraph-developer-proxy/PluginLoader.cs
```diff
@@ -0,0 +1,115 @@
...
+        List<Regex> allWatchedUrls = HandlerConfig.UrlsToWatch.Select(ConvertToRegex).ToList();
+        ISet<Regex> urlsToWatch = allWatchedUrls.ToHashSet();
+        foreach (PluginReference h in config.Handlers) {
+            if (h.Disabled) continue;
+            // Load Handler Assembly if not disabled
+            string? root = Path.GetDirectoryName(typeof(Program).Assembly.Location);
+            if (!string.IsNullOrEmpty(root)) {
+                string pluginLocation = Path.GetFullPath(Path.Combine(root, h.PluginPath.Replace('\\', Path.DirectorySeparatorChar)));
+                PluginLoadContext pluginLoadContext = new PluginLoadContext(pluginLocation);
+                _logger.Log($"Loading from: {pluginLocation}");
```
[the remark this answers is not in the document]
gavinbarron: I think that's a great idea. 

I was deliberately trying to be light touch on refactoring the logging while we solidify that set of functionality.

msgraph-developer-proxy/PluginLoader.cs
```diff
@@ -0,0 +1,115 @@
...
+                Assembly assembly = pluginLoadContext.LoadFromAssemblyName(new AssemblyName(Path.GetFileNameWithoutExtension(pluginLocation)));
+                IEnumerable<Regex>? handlerUrlsList = h.UrlsToWatch?.Select(ConvertToRegex);
+                ISet<Regex>? handlerUrls = null;
+                if (handlerUrlsList is not null) {
+                    handlerUrls = handlerUrlsList.ToHashSet();
+                    allWatchedUrls.AddRange(handlerUrlsList);
```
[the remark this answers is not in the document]
gavinbarron: It needs to append to a top level list so that all watched urls can be registered for SSL decryption in the ProxyEngine

msgraph-developer-proxy/appsettings.json
gavinbarron: Yes, we should, I wanted to have an example of a plugin level set of urlsToWatch for development and testing, I'll move this to the mocksPlugin as we only have graph based mocks in our default set.
Actually I'll disable that plugin in the default set up too.

msgraph-developer-proxy/PluginLoader.cs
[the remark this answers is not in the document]
gavinbarron: To be more clear, my understanding was that a plugin level list of urls is used only for that plugin as a replacement to the default provided at the toplevel

msgraph-proxy-plugins/Guidance/SdkGuidancePlugin.cs
gavinbarron: I've moved this to be triggered by the AfterResponse event. 
The AfterResponse event is raised when calls are passed through to graph and when calls are intercepted during the BeforeRequest event.

msgraph-developer-proxy-plugins/Guidance/SelectGuidancePlugin.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        if (urlsToWatch is null || urlsToWatch.Count == 0) {
+            throw new ArgumentException($"{nameof(urlsToWatch)} cannot be null or empty", nameof(urlsToWatch));
+        }
+
+        _urlsToWatch = urlsToWatch;
+        _logger = context.Logger;
+
+        pluginEvents.BeforeRequest += OnRequest;
```
waldekmastykarz, line 35: Wouldn't it better for this plugin to run on the AfterResponse event so that we show the guidance even if the response has been mocked?

waldekmastykarz (changes requested):

msgraph-developer-proxy/PluginLoader.cs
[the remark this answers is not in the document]
gavinbarron: It doesn't, we make a new HashSet, `defaultUrlsToWatch` from the list loaded out of config on line 29, that is used as a default if there is no value provided for the plugin on line 47, here we're modifying the List which is then used to construct the HashSet that is passed out to be used in the ProxyEngine.

I think that renaming the variables here has helped make the process a bit clearer.

msgraph-developer-proxy-plugins/Guidance/SdkGuidancePlugin.cs
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Graph.DeveloperProxy.Abstractions;
+using System.Text.RegularExpressions;
+using Titanium.Web.Proxy.Http;
+
+namespace Microsoft.Graph.DeveloperProxy.Plugins.Guidance;
+
+public class SdkGuidancePlugin : IProxyPlugin {
```
waldekmastykarz, line 11: One more thought occurred to me: as we're supporting more than Graph, wouldn't it be clearer to name this plugin `GraphSdkGuidancePlugin`, in case we or folks will introduce additional plugins with guidance for other SDKs?

msgraph-developer-proxy-plugins/Guidance/SelectGuidancePlugin.cs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Graph.DeveloperProxy.Abstractions;
+using System.Text.RegularExpressions;
+using Titanium.Web.Proxy.Http;
+
+namespace Microsoft.Graph.DeveloperProxy.Plugins.Guidance;
+
+public class SelectGuidancePlugin : IProxyPlugin {
```
waldekmastykarz, line 11: One more thought occurred to me: as we're supporting more than Graph, wouldn't it be clearer to name this plugin `GraphSelectGuidancePlugin`, in case we or folks will introduce additional plugins with guidance for other APIs?

msgraph-developer-proxy-plugins/MockResponses/MockResponsePlugin.cs
```diff
@@ -0,0 +1,161 @@
...
+    private MockResponseConfiguration _configuration = new();
+    private MockResponsesLoader? _loader = null;
+    private readonly Option<bool> _noMocks;
+    public string Name => nameof(MockResponsePlugin);
+
+    public MockResponsePlugin() {
+        _noMocks = new Option<bool>("--no-mocks", "Disable loading mock requests");
+        _noMocks.AddAlias("-n");
+        _noMocks.ArgumentHelpName = "no mocks";
+        _noMocks.SetDefaultValue(false);
```
gavinbarron, line 38: There should be a command line option to allow developers to pass in the name/path of the mocks file at runtime.

waldekmastykarz: Right now, we're using a fixed name. Shall we add a new issue to make it configurable?

gavinbarron: I already made it configurable in this changeset: https://github.com/microsoftgraph/msgraph-developer-proxy/pull/99/files#diff-d162269ed5127e480a21c801717a78de8a328e3ffc781d0514ba50bcb8c3df04R18

waldekmastykarz (approved):

Issue #68, opened 2022-12-09
Improve readability of console output


waldekmastykarz:
Right now, it's hard to discern the different types of messages in the proxy. What's more, indenting messages makes them even harder to read. We should consider color-coding different types of messages.

Here's our output at the moment:

<img width="975" alt="image" src="https://user-images.githubusercontent.com/11164679/206721857-75055770-b41e-4fd6-b1a9-df3777c9f413.png">

In comparison, here's an example from the [Node proxy](https://github.com/TooTallNate/proxy):

<img width="742" alt="image" src="https://user-images.githubusercontent.com/11164679/206721366-5647e8f7-ef5b-42e4-904f-12fb307885b4.png">

This is debug output where arrows describe different types of messages (proxy response →/request ←, pass-thru response/request ↑).

Let's look at the different types of messages we log and how we could code them to make them easier to follow.

gavinbarron:
I like it! 
We do need to ensure that we use some iconography as well as color for the different types of messages we're outputting for accessibility reasons.
Open question for this, can we find enough text icons in the default character sets across Windows/Linux/Mac or do we need to have the users install a nerdfont?

garrytrinder:
Great suggestion @waldekmastykarz a clearer output will really help.

> can we find enough text icons in the default character sets across Windows/Linux/Mac or do we need to have the users install a nerdfont?

I think we should stick to using text icons and colour coding, but we could also check to see what the emoji support is across the three OS as an alternative to text icons.

waldekmastykarz:
I suspect the lowest common denominator will be cmd on Windows or remote sessions when using the proxy in a container. Let's propose a few designs through a simple console app that we could run across the different terminals to verify it's working as intended. If we choose to use something based on an optional dependency like nerdfont then we should give folks an option to have a readable output without it.

waldekmastykarz:
In the following comments I'll add screenshots of some ideas so far. Use 👍 to vote on your preferred UI

waldekmastykarz:
Option 1

Warp:
<img width="832" alt="image" src="https://user-images.githubusercontent.com/11164679/208642997-292a033a-283d-4385-bacc-29ea98e348d8.png">

Built-in VSCode terminal:
<img width="735" alt="image" src="https://user-images.githubusercontent.com/11164679/208643248-a53d2d85-3279-47ce-ac88-63413fbf67fa.png">

waldekmastykarz:
### Option 2
[...]

PR #114, opened 2023-01-13
Adds console output spec. Related to #68

waldekmastykarz:
Adds console output spec. Related to #68


specs/console-output.md
```diff
@@ -0,0 +1,153 @@
...
+     │ More info at https://learn.microsoft.com/graph/query-parameters#select-parameter
+     └ GET https://graph.microsoft.com/v1.0/me/joinedTeams
+```
+
+## Logging method
+
+To support requirements listed above, and ensure consistent logging from the Proxy's core and its plugins, we'll implement a centralized method, that accepts an array of messages, the type of the message, and context information about the request:
+
+```cs
+interface RequestContext {
```
gavinbarron, line 148: This is a context for logging, and could be in the context of a Request, Response, or General context, suggest we just call it a LoggingContext
```suggestion
interface LoggingContext {
```

specs/console-output.md
```diff
@@ -0,0 +1,153 @@
...
+    public ProxyRequestArgs RequestArgs { get; }
+}
+
+Log(string[] messages, MessageType messageType, RequestContext: context): void
```
gavinbarron, line 152: Do we want to make the Context parameter optional to handle the case where we don't have a SessionEventArgs object to pass?

gavinbarron (changes requested):
Love the overall approach, just need some finetuning on the logging mechanics.

specs/console-output.md
waldekmastykarz: Good point! In that case we'll have a log message that's not related to any request. Not something we use at the moment, but could be useful in the future to avoid breaking the interface.

specs/console-output.md
```diff
@@ -0,0 +1,153 @@
...
+     └ GET https://graph.microsoft.com/v1.0/me/joinedTeams
+```
+
+## Logging method
+
+To support requirements listed above, and ensure consistent logging from the Proxy's core and its plugins, we'll implement a centralized method, that accepts an array of messages, the type of the message, and context information about the request:
+
+```cs
+interface RequestContext {
+    public ProxyRequestArgs RequestArgs { get; }
```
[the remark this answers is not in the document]
waldekmastykarz: The only thing we really need is access to the method and the URL (for now), so reusing `SessionEventArgs` should be sufficient. I don't think we'd need an enum for the context.

gavinbarron (approved):
LGTM

PR #117, opened 2023-01-18
Improves console output readability. Closes #68

### LINKED ISSUE #68: Improve readability of console output

waldekmastykarz:
Improves console output readability. Closes #68

This PR includes:

- implementation of console improvements as outlined in the [spec](https://github.com/microsoftgraph/msgraph-developer-proxy/pull/114)
- introduction of a new `logLevel` config property to control the output level (applies only to non-request processing messages)
- keeps the original logging messages intact to allow logging both proxy- and request-related information


msgraph-developer-proxy-abstractions/ILogger.cs
```diff
@@ -1,10 +1,34 @@
...
+        Session = session;
+    }
+}
+
 public interface ILogger {
-    public void Log(string message);
+    public void LogRequest(string[] message, MessageType messageType, LoggingContext? context = null);
+
+    // Logging methods for non-traffic related messages
+    public void LogInfo(string message);
```
waldekmastykarz, line 31: I kept the original logging methods for logging non-request related messages, like "Press CTRL+C to stop the proxy", etc.

msgraph-developer-proxy-abstractions/ILogger.cs
```diff
@@ -1,10 +1,34 @@
...
+
+    public LoggingContext(SessionEventArgs session)
+    {
+        Session = session;
+    }
+}
+
 public interface ILogger {
-    public void Log(string message);
+    public void LogRequest(string[] message, MessageType messageType, LoggingContext? context = null);
```
waldekmastykarz, line 28: `LogRequest` is used to log decorated messages related to intercepting and processing requests

msgraph-developer-proxy-plugins/RandomErrors/RandomErrorPlugin.cs
```diff
@@ -30,62 +30,62 @@ public class RandomErrorPlugin : IProxyPlugin {
...
-            {
-                "DELETE", new[] {
-                    HttpStatusCode.TooManyRequests,
-                    HttpStatusCode.InternalServerError,
-                    HttpStatusCode.BadGateway,
-                    HttpStatusCode.ServiceUnavailable,
-                    HttpStatusCode.GatewayTimeout,
-                    HttpStatusCode.InsufficientStorage
-                }
+        {
```
waldekmastykarz, line 40: Changed alignment

msgraph-developer-proxy/ConsoleLogger.cs
```diff
@@ -7,30 +7,315 @@ namespace Microsoft.Graph.DeveloperProxy;
...
+
+    public ConsoleLogger(ProxyConfiguration configuration) {
         _color = Console.ForegroundColor;
+        _labelMode = configuration.LabelMode;
+        _logLevel = configuration.LogLevel;
     }
 
-    public void Log(string message) {
+    public void LogInfo(string message) {
+        if (_logLevel > LogLevel.Info) {
```
waldekmastykarz, line 27: In each message I added a comparison with the log level so that users can control which information they want to see.

msgraph-developer-proxy/ConsoleLogger.cs
```diff
@@ -7,30 +7,315 @@ namespace Microsoft.Graph.DeveloperProxy;
...
     }
 
     public void LogWarn(string message) {
+        if (_logLevel > LogLevel.Warn) {
+            return;
+        }
+
         Console.ForegroundColor = ConsoleColor.Yellow;
-        Console.Error.WriteLine($"\tWARNING: {message}");
+        Console.Error.WriteLine($"  WARNING: {message}");
```
waldekmastykarz, line 40: Decreased indentation as it was too big for a terminal with limited width

msgraph-developer-proxy/ConsoleLogger.cs
```diff
@@ -7,30 +7,315 @@ namespace Microsoft.Graph.DeveloperProxy;
...
+        var messageLines = new List<string>(message);
+
+        // add request context information to the message for messages
+        // that are not intercepted requests and have a context
+        if (messageType != MessageType.InterceptedRequest &&
+            context is not null) {
+            messageLines.Add($"{context.Session.HttpClient.Request.Method} {context.Session.HttpClient.Request.Url}");
+        }
+
+        lock (lockObject) {
```
waldekmastykarz, line 74: We need lock to avoid broken rendering:

image

msgraph-developer-proxy/ProxyCommandHandler.cs
```diff
@@ -68,9 +68,9 @@ public class ProxyCommandHandler : ICommandHandler {
 
     }
 
-    private ProxyConfiguration Configuration { get => ConfigurationFactory.Value; }
+    public static ProxyConfiguration Configuration { get => ConfigurationFactory.Value; }
```
waldekmastykarz, line 71: I made the configuration public and static because we need to pass it to the ConsoleLogger which is instantiated before anything else.

msgraph-developer-proxy/ProxyConfiguration.cs
```diff
@@ -1,12 +1,39 @@
...
+public enum LabelMode {
+    [EnumMember(Value = "text")]
+    Text,
+    [EnumMember(Value = "icon")]
+    Icon,
+    [EnumMember(Value = "nerdFont")]
+    NerdFont
+}
+
+public enum LogLevel {
```
waldekmastykarz, line 18: New config property to control the log level

msgraph-developer-proxy-plugins/Guidance/GraphSdkGuidancePlugin.cs
```diff
@@ -42,7 +42,7 @@ public class GraphSdkGuidancePlugin : IProxyPlugin {
             && _urlsToWatch is not null
```
waldekmastykarz, line 42: I renamed this file to align it with the class name

msgraph-developer-proxy-plugins/Guidance/GraphSelectGuidancePlugin.cs
```diff
@@ -38,7 +38,7 @@ public class GraphSelectGuidancePlugin : IProxyPlugin {
     private void AfterResponse(object? sender, ProxyResponseArgs e) {
```
waldekmastykarz, line 38: I renamed this file to align it with the class name

msgraph-developer-proxy/ProxyConfiguration.cs
```diff
@@ -1,12 +1,39 @@
...
+    Error
+}
+
 public class ProxyConfiguration {
     [JsonPropertyName("port")]
     public int Port { get; set; } = 8000;
+    [JsonPropertyName("labelMode")]
+    [JsonConverter(typeof(JsonStringEnumConverter))]
+    public LabelMode LabelMode { get; set; } = LabelMode.Text;
+    [JsonPropertyName("logLevel")]
```
gavinbarron, line 35: Should we expose the LogLevel and LabelMode as command line options too?

msgraph-developer-proxy/ConsoleLogger.cs
```diff
@@ -7,30 +7,315 @@ namespace Microsoft.Graph.DeveloperProxy;
...
+        }
+
         Console.ForegroundColor = ConsoleColor.Yellow;
-        Console.Error.WriteLine($"\tWARNING: {message}");
+        Console.Error.WriteLine($"  WARNING: {message}");
         Console.ForegroundColor = _color;
     }
 
     public void LogError(string message) {
+        if (_logLevel > LogLevel.Error) {
```
gavinbarron, line 45: I like this, even though error is our highest level at present it sets us up well if we need to add something like Critical, nice work.

msgraph-developer-proxy/ProxyConfiguration.cs
waldekmastykarz: logLevel would possibly be useful. I don't think labelMode is something folks would want to change on a per-execution basis.

msgraph-developer-proxy/ProxyHost.cs
```diff
@@ -9,24 +9,29 @@ namespace Microsoft.Graph.DeveloperProxy;
...
     private Option<int> _portOption;
+    private Option<LogLevel> _logLevelOption;
 
     public ProxyHost() {
         _portOption = new Option<int>("--port", "The port for the proxy server to listen on");
         _portOption.AddAlias("-p");
         _portOption.ArgumentHelpName = "port";
         _portOption.SetDefaultValue(8000);
+
+        _logLevelOption = new Option<LogLevel>("--logLevel", "Level of messages to log");
```
gavinbarron, line 20: This should be nullable and not provide a default, otherwise with the current implementation we'll get `LogLevel.Info` if the config is for `warn` and the user doesn't pass a command line option

msgraph-developer-proxy/ProxyCommandHandler.cs
```diff
@@ -31,8 +34,11 @@ public class ProxyCommandHandler : ICommandHandler {
 
     public async Task<int> InvokeAsync(InvocationContext context) {
         int port = context.ParseResult.GetValueForOption(Port);
-        CancellationToken? cancellationToken = (CancellationToken?)context.BindingContext.GetService(typeof(CancellationToken?));
         Configuration.Port = port;
+        LogLevel logLevel = context.ParseResult.GetValueForOption(LogLevel);
+        _logger.LogLevel = logLevel;
```
gavinbarron, line 39: As alluded to in the ProxyHost, we should be null checking logLevel and only assigning here if the user supplies a value.

As I think of it now, port probably needs the same treatment.

msgraph-developer-proxy/responses.json
```diff
@@ -1,104 +1,3 @@
...
-        "mobilePhone": null,
-        "officeLocation": "18/2111",
-        "preferredLanguage": "en-US",
-        "surname": "Vance",
-        "userPrincipalName": "AdeleV@M365x214355.onmicrosoft.com",
-        "id": "87d349ed-44d7-43e1-9a83-5f2406dee5bd"
-      }
-    }
-  ]
+  "responses": []
```
gavinbarron, line 2: Nice! I like this way of giving our users an empty and ready to use set of responses.
Should we output the sample.responses.json file to the output directory so that it's in the zip file users get when they download a release?

gavinbarron (changes requested):
I think we need to fix the handling of the options from the command line

msgraph-developer-proxy/responses.json
waldekmastykarz: Great idea! That way they can copy-paste from us having a starting point 👍

msgraph-developer-proxy/ProxyCommandHandler.cs
waldekmastykarz: Good catch! I'll adjust both.

gavinbarron (approved):
🚀

PR #153, opened 2023-02-01
Fixes new version notification. Closes #150

### LINKED ISSUE #150: [Bug] New Version detection appears broken.
I ran up version 0.3.0 this morning, well after our new v0.4.0 release was out and I was not informed that there was a new version available
image

waldekmastykarz:
Next to fixing the bug, I also centralized current version detection so that we do it consistently.

### DIFF
--- a/msgraph-developer-proxy/ProxyEngine.cs
+++ b/msgraph-developer-proxy/ProxyEngine.cs
@@ -25,15 +25,16 @@ public class ProxyEngine {
     // used for deciding which URLs to decrypt for further inspection
     private ISet<Regex> _hostsToWatch = new HashSet<Regex>();
 
-    private static string __productVersion = string.Empty;
-    private static string _productVersion {
+    private static string _productVersion = string.Empty;
+    public static string ProductVersion {
         get {
-            if (__productVersion == string.Empty) {
-                var fileVersionInfo = FileVersionInfo.GetVersionInfo(AppContext.BaseDirectory);
-                __productVersion = fileVersionInfo?.ProductVersion!;
+            if (_productVersion == string.Empty) {
+                var assemblyPath = typeof(ProxyEngine).Assembly.Location;
+                var fileVersionInfo = FileVersionInfo.GetVersionInfo(assemblyPath);
+                _productVersion = fileVersionInfo?.ProductVersion!;
             }
 
-            return __productVersion;
+            return _productVersion;
         }
     }
 
@@ -197,7 +198,7 @@ public class ProxyEngine {
         }
     }
 
-    private static void AddProxyHeader(Request r) => r.Headers?.AddHeader("Via", $"{r.HttpVersion} graph-proxy/{_productVersion}");
+    private static void AddProxyHeader(Request r) => r.Headers?.AddHeader("Via", $"{r.HttpVersion} graph-proxy/{ProductVersion}");
 
     private bool IsProxiedHost(string hostName) => _hostsToWatch.Any(h => h.IsMatch(hostName));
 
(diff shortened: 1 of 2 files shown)


msgraph-developer-proxy/ProxyEngine.cs
```diff
@@ -25,15 +25,16 @@ public class ProxyEngine {
...
-    private static string __productVersion = string.Empty;
-    private static string _productVersion {
+    private static string _productVersion = string.Empty;
+    public static string ProductVersion {
         get {
-            if (__productVersion == string.Empty) {
-                var fileVersionInfo = FileVersionInfo.GetVersionInfo(AppContext.BaseDirectory);
-                __productVersion = fileVersionInfo?.ProductVersion!;
+            if (_productVersion == string.Empty) {
+                var assemblyPath = typeof(ProxyEngine).Assembly.Location;
```
gavinbarron, line 32: `Assembly.Location` doesn't work in self contained mode and was the source of our issues with plugin loading:
warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.

gavinbarron (changes requested):

msgraph-developer-proxy/ProxyEngine.cs
waldekmastykarz: Right. We need to find a way to get the executable name as well on top of its location.

Is there btw a way to use the same type of build locally as we use in the workflow to avoid different behaviors and catch such discrepancies earlier?

gavinbarron: from the CLI absolutely! which means you can probably configure it in VSCode. This is building a release build, but changing the -c option to Debug should work just fine

```bash
 dotnet build ./msgraph-developer-proxy/msgraph-developer-proxy.csproj -c Release -r osx-x64 --self-contained
 dotnet build ./msgraph-developer-proxy-plugins/msgraph-developer-proxy-plugins.csproj -c Release -r osx-x64 --no-self-contained
```

gavinbarron: Regarding the file path to the executable, perhaps this helps: https://github.com/dotnet/runtime/issues/13051#issuecomment-510267727

waldekmastykarz: It's odd, but when I use the build statements you shared above locally, I can get the assembly path just fine:

image

image

I tried both Release and Debug mode to see if it makes a difference, but in both cases the path is displayed just fine. What am I missing?

gavinbarron: That is really odd. I'll take a look and see if I can reproduce it.

gavinbarron: If we add 
```
<PublishSingleFile>true</PublishSingleFile>
```

to `msgraph-developer-proxy.csproj` as a child of the first `<PropertyGroup>` element then in full VS we'll get better hints that about the Assembly.Location issue:
image

gavinbarron: How about we make this change?

```suggestion
                var assemblyPath = Process.GetCurrentProcess()?.MainModule?.FileName ?? typeof(ProxyEngine).Assembly.Location;
```

waldekmastykarz:
Hey @gavinbarron, now that we've got all code changes, would you mind having another look?

gavinbarron (approved):
🚀

Issue #93, opened 2022-12-22
Allow recording requests


waldekmastykarz:
Users should be able to put the Proxy in recording mode where it will capture requests. After recording is stopped, the Proxy will be able to execute an action on the recorded requests, like analyzing it for minimal permissions, building a report, etc.

Depends on #58

PR #155, opened 2023-02-01
Adds recording mode spec. Related to #93

waldekmastykarz:
Adds recording mode spec. Related to #93

### DIFF
--- a/specs/recording-requests.md
+++ b/specs/recording-requests.md
@@ -10,7 +10,7 @@ Recording requests allows Developer Proxy to capture requests and related messag
 
 ## Implementation
 
-Developer Proxy can be put in recording mode either by using the `--record` command line option or by pressing `r` after starting the proxy. Recording can be stopped by pressing `s` while running the proxy or by closing the proxy using `CTRL+C`. When the recording is stopped while proxy is running, users can start a new recording session by pressing `r` again.
+Developer Proxy can be put in recording mode either by using the `--record` command line option or by pressing `r` after starting the proxy. Recording can be stopped by pressing `s` while running the proxy or by closing the proxy using `CTRL+C`. While recording is stopped and the proxy is running, users can start a new recording session by pressing `r` again.
 
 While recording, Developer Proxy will show and indicator in the terminal window. The indicator will be displayed in the top right corner of the window and will be hidden if the proxy is not recording.
 
--- a/specs/recording-requests.md
+++ b/specs/recording-requests.md
@@ -12,7 +12,7 @@ Recording requests allows Developer Proxy to capture requests and related messag
 
 Developer Proxy can be put in recording mode either by using the `--record` command line option or by pressing `r` after starting the proxy. Recording can be stopped by pressing `s` while running the proxy or by closing the proxy using `CTRL+C`. While recording is stopped and the proxy is running, users can start a new recording session by pressing `r` again.
 
-While recording, Developer Proxy will show and indicator in the terminal window. The indicator will be displayed in the top right corner of the window and will be hidden if the proxy is not recording.
+While recording, Developer Proxy will show an indicator in the terminal window. The indicator will be displayed in the top right corner of the window and will be hidden if the proxy is not recording.
 
 Proxy is recording requests by subscribing to a `LogRequest` event which is raised each time proxy logs a request-related log message. The event contains information about the logged message.
 
--- a/specs/recording-requests.md
+++ b/specs/recording-requests.md
@@ -1,6 +1,6 @@
 # Recording requests
 
-Recording requests allows Developer Proxy to capture requests and related messages that were intercepted by the proxy while in record mode. The captured requests are passed for further processing to the Developer Proxy plugins. This feature is foundational for scenarios such as permission analysis, building tailored SDKs, audit reports, etc.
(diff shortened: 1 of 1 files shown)


specs/recording-requests.md
```diff
@@ -0,0 +1,21 @@
...
+
+## History
+
+| Version | Date | Comments | Author |
+| ------- | ---- | -------- | ------ |
+| 1.0 | 2023-02-01 | Initial specifications | @waldekmastykarz |
+
+## Implementation
+
+Developer Proxy can be put in recording mode either by using the `--record` command line option or by pressing `r` after starting the proxy. Recording can be stopped by pressing `s` while running the proxy or by closing the proxy using `CTRL+C`. When the recording is stopped while proxy is running, users can start a new recording session by pressing `r` again.
```
gavinbarron, line 13:
```suggestion
Developer Proxy can be put in recording mode either by using the `--record` command line option or by pressing `r` after starting the proxy. Recording can be stopped by pressing `s` while running the proxy or by closing the proxy using `CTRL+C`. While recording is stopped and the proxy is running, users can start a new recording session by pressing `r` again.
```

specs/recording-requests.md
```diff
@@ -0,0 +1,21 @@
...
+
+While recording, Developer Proxy will show and indicator in the terminal window. The indicator will be displayed in the top right corner of the window and will be hidden if the proxy is not recording.
```
gavinbarron, line 15:
```suggestion
While recording, Developer Proxy will show an indicator in the terminal window. The indicator will be displayed in the top right corner of the window and will be hidden if the proxy is not recording.
```

specs/recording-requests.md
```diff
@@ -0,0 +1,21 @@
+# Recording requests
+
+Recording requests allows Developer Proxy to capture requests and related messages that were intercepted by the proxy while in record mode. The captured requests are passed for further processing to the Developer Proxy plugins. This feature is foundational for scenarios such as permission analysis, building tailored SDKs, audit reports, etc.
```
gavinbarron, line 3: Recording is such an overloaded term, I got this idea confused with the idea of recording requests and responses to generate a mock responses file.

Perhaps we should consider being more explicit in the title and opening that we're recording the logging data?

specs/recording-requests.md
```diff
@@ -0,0 +1,21 @@
...
+| ------- | ---- | -------- | ------ |
+| 1.0 | 2023-02-01 | Initial specifications | @waldekmastykarz |
+
+## Implementation
+
+Developer Proxy can be put in recording mode either by using the `--record` command line option or by pressing `r` after starting the proxy. Recording can be stopped by pressing `s` while running the proxy or by closing the proxy using `CTRL+C`. When the recording is stopped while proxy is running, users can start a new recording session by pressing `r` again.
+
+While recording, Developer Proxy will show and indicator in the terminal window. The indicator will be displayed in the top right corner of the window and will be hidden if the proxy is not recording.
+
+Proxy is recording requests by subscribing to a `LogRequest` event which is raised each time proxy logs a request-related log message. The event contains information about the logged message.
```
gavinbarron, line 17: Is this event to be raised on each call to `ILoger.LogRequest`?
Does it happen before the logging happens or after?

specs/recording-requests.md
```diff
@@ -0,0 +1,21 @@
...
+
+After the recording is stopped, Developer Proxy will raise the `RecordingStopped` event. Plugins can subscribe to this event to process the captured requests.
```
gavinbarron, line 19: If I'm understanding this idea correctly we could have a new kind of plugin, say a `SessionSummaryPlugin` that on `RecordingStopped` would receive all of the collected logged request data and messages. This plugin could then generate some output based on the data it receives.

specs/recording-requests.md
waldekmastykarz: This would be totally possible with a dedicated recording plugin! 😀 

Would capturing be clearer?

specs/recording-requests.md
waldekmastykarz: Yes, it's this event. I suggest we raise the event after the message was logged. It aligns more with the "Logged" aspect (past tense).

specs/recording-requests.md
waldekmastykarz: Correct, so this gives us a lot of flexibility when it comes to processing a set of requests, like summaries, analysis or passing on to other tools like Kiota.

specs/recording-requests.md
gavinbarron: Perhaps a slight tweak to the event name `RequestLogged` gets this to the place we want it.

IMO, for the moment we leave it and address when/if we focus on custom loggers.

specs/recording-requests.md
gavinbarron: Perhaps we call this Record request logging, something like:
```suggestion
Record request logging allows Developer Proxy to capture request logging activities that occurred while the proxy was in record mode. The captured request logging items are passed for further processing to the Developer Proxy plugins. This feature is foundational for scenarios such as permission analysis, building tailored SDKs, audit reports, etc.
```

specs/recording-requests.md
```diff
@@ -0,0 +1,21 @@
+# Recording requests
```
gavinbarron, line 1:
```suggestion
# Record request logging
```

gavinbarron (approved):

PR #174, opened 2023-02-06
Adds odata paging guidance. Closes #151

### LINKED ISSUE #151: Guidance: Is my app correctly supporting paging guidance
Offer guidance that verifies that apps properly handle retrieving multiple pages of data by using OData next page links rather than building links themselves. Include handling `$skip` values and cursors.

waldekmastykarz:
Adds odata paging guidance. Closes #151 

I also found a bug in how we resolve hosts to watch, where hosts with wildcards like `*.sharepoint.*` would be incorrectly processed for RegEx and invalid.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,13 +10,13 @@
       "request": "launch",
       "preLaunchTask": "build",
       // If you have changed target frameworks, make sure to update the program path.
-      "program": "${workspaceFolder}/msgraph-developer-proxy/bin/Debug/net6.0/msgraph-developer-proxy.dll",
+      "program": "${workspaceFolder}/msgraph-developer-proxy/bin/Debug/net6.0/mgdp.dll",
       "args": [],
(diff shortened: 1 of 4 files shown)


sebastienlevert:
First thing that comes to mind... Should we have a series of aka.ms links that we control to be able to change location of guidance docs if they ever change without re-releasing a new version of the proxy? That should apply to all links we link to.

waldekmastykarz:
Good idea! 👍

waldekmastykarz:
> First thing that comes to mind... Should we have a series of aka.ms links that we control to be able to change location of guidance docs if they ever change without re-releasing a new version of the proxy? That should apply to all links we link to.

Thinking of it: shall we take it on a separate issue, so that we can consistently update all links in one go?

sebastienlevert:
> Thinking of it: shall we take it on a separate issue, so that we can consistently update all links in one go?

I think so!

gavinbarron (approved):

Issue #152, opened 2023-02-01
Provide execution summary


waldekmastykarz:
Provide a summary of captured requests and their related messages. Depends on #93

waldekmastykarz:
Let's wait with writing this spec until we clarified spec for #93 to avoid unnecessary rework.

PR #177, opened 2023-02-06
Adds Execution summary spec. Related to #152

waldekmastykarz:
Adds Execution summary spec. Related to #152

@gavinbarron @sebastienlevert @garrytrinder I'd appreciate your review


specs/execution-summary.md
```diff
@@ -0,0 +1,178 @@
...
+Execution summary contains the following information:
+
+- Summary title (fixed)
+- Date and time of the recording session
+- Groups as specified by the `groupBy` configuration option:
+  - if grouped by URL:
+    - List of requests captured during the recording session sorted alphabetically. For each request:
+      - Request method
+      - Request URL
+      - List of unique request log messages for this request issued by the proxy, grouped and sorted by severity. Excluding the `InterceptedRequest` and `PassedThrough` severity.
```
sebastienlevert, line 35: Can we also have the number of instances for this log message? I feel it would be relevant to know that we had 22 WARNINGS on the /me/directReports for instance.

specs/execution-summary.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### https://graph.microsoft.com/v1.0/me/*
+
+- GET https://graph.microsoft.com/v1.0/me/messages
+
+### Tips
+
+#### To handle API errors more easily, use the Graph SDK. More info at https://aka.ms/move-to-graph-js-sdk
+
+- GET https://graph.microsoft.com/v1.0/me
```
sebastienlevert, line 153: Same here, per method / URL we should know the # of unique instances of this TIP (or warning, failure, etc.)

sebastienlevert (changes requested):

specs/execution-summary.md
waldekmastykarz: Good idea. Shall we add the count after each message, eg.

```md
### GET https://graph.microsoft.com/v1.0/me

#### Warning

- To improve performance of your application, use the $select parameter. ore info at https://learn.microsoft.com/graph/query-parameters#select-parameter (22)
```

specs/execution-summary.md
waldekmastykarz: Would the same approach, with the count of instances displayed between `()`, work here as well?

```md
## Message types

### Requests intercepted

- GET https://graph.microsoft.com/v1.0/me (10)
- GET https://graph.microsoft.com/v1.0/me/messages (15)
- GET https://graph.microsoft.com/v1.0/me/users (12)
```

specs/execution-summary.md
sebastienlevert:
```suggestion
      - (22) To improve performance of your application, use the $select parameter. ore info at https://learn.microsoft.com/graph/query-parameters#select-parameter (22)
```

specs/execution-summary.md
sebastienlevert: I'd prepend it for context.

specs/execution-summary.md
waldekmastykarz: Would you display the number of occurrences both before and after? 😕

sebastienlevert: Bad copy/paste haha! Just before!

appieschot: When grouping per URL it might make sense to provide CSV exports as well. Had a internal discussion with some testers who considered testing the application for extended periods op time (so lets say you have a complex process that consists of 15 steps you might want to know what happens in each step when 10% of our calls fail). Using the proxy you facilitate that scenario but with the report a CSV with all failed requests that contains `TimeStamp, Request Method, Request URL, Error Message` devs can map errors to what might be logged in App Insights to get even more details and do better analysis. I assume it should be a specific configuration option and the CSV should live next to the markdown output.

waldekmastykarz: Would the CSV be just another output form next to console or would it be a summary? In other words: would you expect to see in the CSV all occurrences of the same error or only the unique occurrences?

gavinbarron (approved):

sebastienlevert (approved):

PR #179, opened 2023-02-07
Fixes issue with mock responses not being set. Closes #164

### LINKED ISSUE #164: [Bug🐞/or I am doing something completely wrong🤔] - mocking response is not send
I have a test WPF app that sends request to `https://graph.microsoft.com/v1.0/me`. 
When I use chaos mode and a failure is generated I get the correct behavior, which means I get the error in the response 
image

When I use the same mocking the response to return `500` it seems the request is caught but I as a response I don't get the error from the proxy but I get response from graph API
image

this is my `mock.json` file:
```
{
  "responses": [
    {
      "url": "https://graph.microsoft.com/v1.0/me",
      "method": "GET",
      "responseCode": 500
    },
    {
      "url": "https://customapiendpoint/test/gettest",
      "method": "GET",
      "responseCode": 500
    }
  ]
}
```

Am I missing something or is it a bug?
TBH the same kind of scenario worked fine in v0.3 meaning I got an error in the response when mocking.

waldekmastykarz:
Fixes issue with mock responses not being set. Closes #164

### DIFF
--- a/msgraph-developer-proxy-plugins/MockResponses/MockResponsePlugin.cs
+++ b/msgraph-developer-proxy-plugins/MockResponses/MockResponsePlugin.cs
@@ -162,8 +162,8 @@ public class MockResponsePlugin : BaseProxyPlugin {
             else {
                 body = bodyString;
             }
-            e.GenericResponse(body ?? string.Empty, statusCode, headers);
         }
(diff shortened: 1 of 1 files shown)


gavinbarron (approved):
Ahhh, the case where we don't have a response body from the mock file is a valid one, oops...

PR #185, opened 2023-02-15
Update links to use aka.ms. Closes #180

### LINKED ISSUE #180: Define aka.ms links for all URLs used in Developer Proxy
For all URLs that we use in the Developer Proxy, let's use aka.ms links so that we can change them if necessary without breaking old versions of Proxy.

--- original

First thing that comes to mind... Should we have a series of aka.ms links that we control to be able to change location of guidance docs if they ever change without re-releasing a new version of the proxy? That should apply to all links we link to.

_Originally posted by @sebastienlevert in https://github.com/microsoftgraph/msgraph-developer-proxy/issues/174#issuecomment-1419354782_

garrytrinder:
Closes #180

### DIFF
--- a/msgraph-developer-proxy-plugins/Guidance/GraphBetaSupportGuidancePlugin.cs
+++ b/msgraph-developer-proxy-plugins/Guidance/GraphBetaSupportGuidancePlugin.cs
@@ -28,6 +28,6 @@ public class GraphBetaSupportGuidancePlugin : BaseProxyPlugin {
             _logger?.LogRequest(BuildBetaSupportMessage(request), MessageType.Warning, new LoggingContext(e.Session));
     }
 
-    private static string GetBetaSupportGuidanceUrl() => "https://graph/proxy/guidance/beta";
+    private static string GetBetaSupportGuidanceUrl() => "https://aka.ms/graph/proxy/guidance/beta-support";
     private static string[] BuildBetaSupportMessage(Request r) => new[] { $"Don't use beta APIs in production because they can change or be deprecated.", $"More info at {GetBetaSupportGuidanceUrl()}" };
 }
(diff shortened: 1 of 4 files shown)


msgraph-developer-proxy-plugins/Guidance/GraphBetaSupportGuidancePlugin.cs
```diff
@@ -28,6 +28,6 @@ public class GraphBetaSupportGuidancePlugin : BaseProxyPlugin {
             _logger?.LogRequest(BuildBetaSupportMessage(request), MessageType.Warning, new LoggingContext(e.Session));
     }
 
-    private static string GetBetaSupportGuidanceUrl() => "https://learn.microsoft.com/graph/versioning-and-support#beta-version";
+    private static string GetBetaSupportGuidanceUrl() => "https://graph/proxy/guidance/beta";
```
waldekmastykarz, line 31: I think we're missing aka.ms here

waldekmastykarz, line 31: Shall we make it more specific, like `beta-support` or something like that in case we'll introduce more beta-related guidance, like moving beta to v1.0 endpoints when available?

waldekmastykarz (changes requested):
Nice! Let's do a few adjustments before we merge @garrytrinder

msgraph-developer-proxy-plugins/Guidance/GraphBetaSupportGuidancePlugin.cs
garrytrinder: 🤦‍♂️ we are indeed, good catch.

garrytrinder: Good suggestion 👍

waldekmastykarz (approved):

Issue #210, opened 2023-02-28
Add rate limiting guidance


waldekmastykarz:
Add rate limiting guidance so that developers can test that their apps properly back off when they're about to exceed resource limits for the current timeframe.

The guidance should work with any API

PR #221, opened 2023-03-08
Adds the Rate limiting guidance spec

waldekmastykarz:
Adds the Rate limiting guidance spec. Related to #210


specs/rate-limiting.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Configuration
+
+Setting|Description
+-------|-----------
+`headerLimit`|Name of the `RateLimit-Limit` header returned by the API. Default: `RateLimit-Limit`
+`headerRemaining`|Name of the `RateLimit-Remaining` header returned by the API. Default: `RateLimit-Remaining`
+`headerReset`|Name of the `RateLimit-Reset` header returned by the API. Default: `RateLimit-Reset`
+`headerRetryAfter`|Name of the `Retry-After` header returned by the API. Default: `Retry-After`
+`costPerRequest`|Number of resources consumed by a single request. Default: `1`
```
jansenbe, line 21: Assume you mean resource units here? Let's make that 2 as generic average, in line with the docs https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online#application-throttling.

specs/rate-limiting.md
```diff
@@ -0,0 +1,41 @@
...
+`resetTimeWindowSeconds`|Number of seconds before the resource counter is reset. Default: `60`
+`warningThresholdPercent`|Percentage of remaining resources at which the warning should be displayed. Default: `80`
+`rateLimit`|Number of resources available in the time window. Default: `120`
```
jansenbe, line 24: Where does 120 come from? The lowest value 1200 resource units per minute

waldekmastykarz: This is just a random value indeed meant to help folks verify that their code properly handles backing off without having to issue an excessive number of requests first. We can include guidance in our docs to the correct documentation in case folks want to see the behavior closer to Microsoft Graph.

sebastienlevert:
Couple of comments: 

- Should we think of adding a guidance for developers to let them know that the endpoint they are calling supports RateLimiting?  
- Should we generate randomly some high cost and low cost requests to replicate real-life behaviors?
- Can we assume that all /sites and /drives support it? Should we emit these only for these endpoints by default in the plugin config? It would be to the developer to "open up" to other endpoints.
- We don't have rate limiting support in SDKs today, we should add a work item over there (for kiota-generated clients) but before, write a guidance for all languages (or at least JS and .NET) on how to react to it.

waldekmastykarz:
@sebastienlevert

> Should we think of adding a guidance for developers to let them know that the endpoint they are calling supports RateLimiting?

That would be a nice separate plugin for Microsoft Graph

> Should we generate randomly some high cost and low cost requests to replicate real-life behaviors?

This would be a good next step. The basic setup described in the current spec allows customer to answer the question: "Can my app properly back off to avoid being throttled and maximize throughput?" By mapping types of requests to request costs on Microsoft Graph we'd help them understand how they application will behave in production. I suggest we take this separately.

> We don't have rate limiting support in SDKs today, we should add a work item over there (for kiota-generated clients) but before, write a guidance for all languages (or at least JS and .NET) on how to react to it.

The tricky part is implementing the logic to back off. The SDK doesn't have the context of the app, it doesn't know how many calls are there left and how many calls are being executed in parallel. As a result, it could start backing off even though the remaining number of requests would fit just fine in the current resource window. While we could offer something simple and generic like _start backing off when the app reaches 80% or resources_ I wonder how valuable this would be in real life.

specs/rate-limiting.md
```diff
@@ -0,0 +1,43 @@
...
+`retryAfterSeconds`|Number of seconds to wait before retrying the request. Default: `5`
+
+## Implementation
+
+Rate limiting guidance will be implemented as a plugin, subscribed to the `BeforeRequest` event.
+
+The plugin starts with the total number of resources available and the time of next reset as configured in the `rateLimit` and `resetTimeWindowSeconds` settings.
```
gavinbarron, line 31: Is this reset calculated using a sliding or rolling window? I'm a little unclear based on what we have here and the documentation, for throttling in SPO it could be either. It would be good to be explicit here.

specs/rate-limiting.md
```diff
@@ -0,0 +1,43 @@
...
+`headerLimit`|Name of the `RateLimit-Limit` header returned by the API. Default: `RateLimit-Limit`
+`headerRemaining`|Name of the `RateLimit-Remaining` header returned by the API. Default: `RateLimit-Remaining`
+`headerReset`|Name of the `RateLimit-Reset` header returned by the API. Default: `RateLimit-Reset`
+`headerRetryAfter`|Name of the `Retry-After` header returned by the API. Default: `Retry-After`
+`costPerRequest`|Number of resources consumed by a single request. Default: `2`
+`resetTimeWindowSeconds`|Number of seconds before the resource counter is reset. Default: `60`
+`warningThresholdPercent`|Percentage of remaining resources at which the warning should be displayed. Default: `80`
+`rateLimit`|Number of resources available in the time window. Default: `120`
+`retryAfterSeconds`|Number of seconds to wait before retrying the request. Default: `5`
+
```
gavinbarron, line 26: Can this plugin define its own set of URLs to watch or must it follow the overall set?

specs/rate-limiting.md
```diff
@@ -0,0 +1,43 @@
...
+When the number of resources left drops to zero, the plugin will start returning `429 Too Many Requests` response and log a FAIL request log message (the application failed to back off in time to avoid throttling). The plugin will also return the number of seconds to wait before retrying the request as configured in the `retryAfterSeconds` setting. This information will be sent in the header configured in the `headerRetryAfter` setting.
+
+If the application tries calling the API again before the time specified in the `Retry-After` header, the API will return `429 Too Many Requests` again and will log a FAIL request log message (the application failed to wait the specified amount of time before calling the API again).
+
+When the time of next reset is reached, the plugin will reset the total number of resources available to the value of the `rateLimit` setting.
```
gavinbarron, line 39: This strongly suggests a rolling rate limiting window. Might be good to get confirmation from SPO that is their approach internally?

specs/rate-limiting.md
waldekmastykarz: I was under the impression that SPO uses a sliding window because it communicates time to next reset.

specs/rate-limiting.md
waldekmastykarz: I assumed that configuring own set of URL is a common base functionality of all plugins which is why I haven't mentioned it here explicitly.

specs/rate-limiting.md
waldekmastykarz: Here's the sample project @jansenbe put together to demonstrate and test the functionality: https://github.com/OneDrive/samples/tree/master/scenarios/throttling-ratelimit-handling

gavinbarron (approved):

PR #224, opened 2023-03-09
Changes BeforeResponse to async event

waldekmastykarz:
Changes BeforeResponse to async event to fix the issue with OData paging guidance

### DIFF
--- a/msgraph-developer-proxy-abstractions/FuncExtensions.cs
+++ b/msgraph-developer-proxy-abstractions/FuncExtensions.cs
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// from: https://github.com/justcoding121/titanium-web-proxy/blob/902504a324425e4e49fc5ba604c2b7fa172e68ce/src/Titanium.Web.Proxy/Extensions/FuncExtensions.cs
+
+using Titanium.Web.Proxy;
+using Titanium.Web.Proxy.EventArguments;
+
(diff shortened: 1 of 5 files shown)


gavinbarron:
I gave it a run and it seems to be working correctly now.  Great work as usual!

I went on a little research trip to see what I could learn about mechanisms we can have for handling async events, I found a couple of very interesting articles.
The first goes deep into how async in dotnet uses SynchronizationContext and state machines, super interesting stuff: https://olegkarasik.wordpress.com/2019/04/16/code-tip-how-to-work-with-asynchronous-event-handlers-in-c/

The second looks a lot like the approach we have here: https://medium.com/@a.lyskawa/the-hitchhiker-guide-to-asynchronous-events-in-c-e9840109fb53
I suspect the second approach is a little lightweight and doesn't have quite as robust error handling or handling for mixed sync and async handlers. 

That said I think that we can likely refactor the AsyncEvents such that we don't need to add more points where Titanium bleeds out into the plugins. I'm also comfortable with accepting that we have a bit of abstraction leakage at the moment and revisiting it in the future.

waldekmastykarz:
Thanks for the additional references. As we're looking into refactoring in the future anyway, I suggest that we proceed with using async events from Titanium for now and decide on the final approach before GA.

gavinbarron:
I agree

gavinbarron:
This looks really good, the only question I have is do we want to add something like `await Task.CompletedTask;` or wrap the logic in the Async event receivers in `await Task.Run` to eliminate the compiler warnings about async methods without awaits?

gavinbarron (approved):

PR #229, opened 2023-03-13
Adds support for specifying the config file name. Closes #228

### LINKED ISSUE #228: Allow specifying config file name
Right now we're using a config file with a fixed name: appsettings.json. As we're expanding proxy's features, it would be beneficial to allow defining presets for different APIs and scenarios, eg: testing Microsoft Graph random errors, testing contoso app scalability on M365, testing SPO, testing Open AI API, testing contoso API, etc. While you could create a separate file for each of these scenarios already now, in order to use them, you'd have to rename appsettings.json to something else, then rename the specific preset to appsettings.json. When you're done, you'd have to revert your changes and rename the files back. It would be much more convenient if you could start proxy with a specific config file, eg: `mgdp --config my-preset.json`, and use appsettings.json as the default settings when no config has been specified.

waldekmastykarz:
Adds support for specifying the config file name. Closes #228

### DIFF
--- a/msgraph-developer-proxy/ProxyHost.cs
+++ b/msgraph-developer-proxy/ProxyHost.cs
@@ -11,6 +11,43 @@ internal class ProxyHost {
     private Option<int?> _portOption;
     private Option<LogLevel?> _logLevelOption;
     private Option<bool?> _recordOption;
+    private static Option<string?>? _configFileOption;
+
+    private static bool _configFileResolved = false;
+    private static string _configFile = "appsettings.json";
+    public static string ConfigFile {
+        get {
+            if (_configFileResolved) {
+                return _configFile;
+            }
+            
+            if (_configFileOption is null) {
+                _configFileOption = new Option<string?>("--config-file", "The path to the configuration file");
+                _configFileOption.AddAlias("-c");
+                _configFileOption.ArgumentHelpName = "configFile";
+                _configFileOption.AddValidator(input => {
+                    var filePath = input.Tokens.First().Value;
+                    if (String.IsNullOrEmpty(filePath)) {
+                        return;
+                    }
+
+                    if (!File.Exists(filePath)) {
+                        input.ErrorMessage = $"File {filePath} does not exist";
+                    }
+                });
+            }
+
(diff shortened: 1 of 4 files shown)


msgraph-developer-proxy/ProxyHost.cs
```diff
@@ -11,6 +11,43 @@ internal class ProxyHost {
     private Option<int?> _portOption;
     private Option<LogLevel?> _logLevelOption;
     private Option<bool?> _recordOption;
+    private static Option<string?>? _configFileOption;
+
+    private static bool _configFileResolved = false;
+    private static string _configFile = "appsettings.json";
+    public static string ConfigFile {
```
waldekmastykarz, line 18: Implemented this as a static with custom option parsing, because the value of this option needs to be available before we invoke the RootCommand which is typically responsible for parsing args.

msgraph-developer-proxy/ProxyHost.cs
```diff
@@ -32,7 +69,8 @@ internal class ProxyHost {
         var command = new RootCommand {
             _portOption,
             _logLevelOption,
-            _recordOption
+            _recordOption,
+            _configFileOption!
```
waldekmastykarz, line 73: Added the option to the root command as well, so that it shows up in help

waldekmastykarz:
@garrytrinder, after we merge this PR, let's ensure to add this new feature to our docs

msgraph-developer-proxy/ProxyHost.cs
gavinbarron: Let's add a comment to explain that `_configFIleOption` gets set during the call to load `ProxyCommandHandler.Configuration` in the main program before the call to create the root command happens.
There's a few levels of indirection here and this could easily get missed during a refactor of the startup process

gavinbarron:
Very nice work!

gavinbarron (approved):

msgraph-developer-proxy/ProxyHost.cs
waldekmastykarz: Great point. Definitely makes sense to take note of that assumption here.

PR #230, opened 2023-03-13
Intercept requests from a specific process. Closes #227

### LINKED ISSUE #227: Feature suggestion: Target only single app
Whenever I'm using the developer proxy soo many apps start to break, - for instance I my teams client sort of breaks and I don't always get messages, and while I don't mind, my boss does for some reason!

A way to attach the proxy to only a single service would be awesome!

waldekmastykarz:
Intercept requests from a specific process. Closes #227


gavinbarron:
I tested this on Windows and it works well, I used `--watchpids` to intercept requests from an instance of PostMan and `--watch-process-names`  to intercept request from msedge. Both sets of requests to graph were shown in the logs. 

Restarting PostMan resulted in requests not being intercepted as they were from a new process and hence processId.

I will say that getting the specific processId for edge proved rather tricky and I actually gave up and used the process name as Edge (And Chromium in general) spawns so many processes. Do you have good trick for isolating the processId that should be used when looking to bind a single Chromium instance?

Tested the validation of the process id and that's working nicely out of the box:
image

gavinbarron (approved):
👍

sebastienlevert:
This is crazy cool!

PR #232, opened 2023-03-15
Adds the rate limiting plugin. Closes #210

### LINKED ISSUE #210: Add rate limiting guidance

waldekmastykarz:
Adds the rate limiting plugin. Closes #210

Plugin is disabled by default because it works only when mocking and random errors are disabled. We can provide specific presets after we've done #228

### DIFF
--- a/msgraph-developer-proxy-plugins/Behavior/RateLimitingPlugin.cs
+++ b/msgraph-developer-proxy-plugins/Behavior/RateLimitingPlugin.cs
@@ -1,0 +1,212 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Graph.DeveloperProxy.Abstractions;
+using System.Net;
+using System.Text.Json;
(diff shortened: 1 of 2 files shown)


waldekmastykarz:
@garrytrinder, after we merge this, let's ensure we add it to the docs

waldekmastykarz:
> I'd love if we make this default for Graph endpoints. ODSP is using those all over the place and I feel this drives a lot of value over there. I'd love a developer to go on and test their code without having to change the config for this scenario.

Given that rate limiting works when chaos is disabled, I suggest that we start thinking about including some scenario-based presets that are ready to use. That way, we could offer preconfigured settings for the different APIs and scenarios that we want to test, without offloading it all to our customers.

gavinbarron (approved):
LGTM. Do you want to hold merging this until #224 is done and this can be refactored to use the AsyncEvents pattern?

waldekmastykarz:
Good idea!

PR #242, opened 2023-04-07
Fixes issue with throttling. Closes #241

### LINKED ISSUE #241: [BUG]: Requests are always passed through when clients wait based on Retry-Header value returned
This bug affects both the `GenericRandomErrorPlugin` and `GraphRandomErrorPlugin` when simulating throttling.

If the first request is throttled and the app waits for the `Retry-After` period before sending the next request, the request is always passed through.

The proxy should either pass through the request or fail the request with a random error based on the configured failure rate.

garrytrinder:
Ensures that when a client waits for the given Retry-After period that the proxy can return a random error response as well as passing through the request. Closes #241

### DIFF
--- a/msgraph-developer-proxy-plugins/RandomErrors/GenericRandomErrorPlugin.cs
+++ b/msgraph-developer-proxy-plugins/RandomErrors/GenericRandomErrorPlugin.cs
@@ -52,15 +52,8 @@ public class GenericRandomErrorPlugin : BaseProxyPlugin {
                 return GenericRandomErrorFailMode.Throttled;
             }
             else {
-                // request was sent after retryAfterDate so we don't force fail
-                // work out if we should fail the next request or pass it through
-                var result = _random.Next(1, 100) <= _configuration.Rate ? GenericRandomErrorFailMode.Random : GenericRandomErrorFailMode.PassThru;
-                // clean up expired throttled request if request to be passed through
-                if (result == GenericRandomErrorFailMode.PassThru)
-                {
-                    _throttledRequests.Remove(key);
-                }
-                return result;
+                // clean up expired throttled request and ensure that this request is passed through.
+                _throttledRequests.Remove(key);
             }
         }
         return _random.Next(1, 100) <= _configuration.Rate ? GenericRandomErrorFailMode.Random : GenericRandomErrorFailMode.PassThru;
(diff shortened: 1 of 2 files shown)


msgraph-developer-proxy-plugins/RandomErrors/GenericRandomErrorPlugin.cs
```diff
@@ -52,9 +52,15 @@ public class GenericRandomErrorPlugin : BaseProxyPlugin {
...
-                return GenericRandomErrorFailMode.PassThru;
+                // request was sent after retryAfterDate so we don't force fail
+                // work out if we should fail the next request or pass it through
+                var result = _random.Next(1, 100) <= _configuration.Rate ? GenericRandomErrorFailMode.Random : GenericRandomErrorFailMode.PassThru;
+                // clean up expired throttled request if request to be passed through
+                if (result == GenericRandomErrorFailMode.PassThru)
+                {
+                    _throttledRequests.Remove(key);
+                }
+                return result;
```
waldekmastykarz, line 63:
```suggestion
                // clean up expired throttled request and ensure that this request is passed through.
                _throttledRequests.Remove(key);
```

msgraph-developer-proxy-plugins/RandomErrors/GraphRandomErrorPlugin.cs
```diff
@@ -119,9 +119,15 @@ public class GraphRandomErrorPlugin : BaseProxyPlugin {
...
-                return GraphRandomErrorFailMode.PassThru;
+                // request was sent after retryAfterDate so we don't force fail
+                // work out if we should fail the next request or pass it through
+                var result = _random.Next(1, 100) <= _configuration.Rate ? GraphRandomErrorFailMode.Random : GraphRandomErrorFailMode.PassThru;
+                // clean up expired throttled request if request to be passed through
+                if (result == GraphRandomErrorFailMode.PassThru)
+                {
+                    _throttledRequests.Remove(key);
+                }
+                return result;
```
waldekmastykarz, line 130:
```suggestion
                // clean up expired throttled request and ensure that this request is passed through.
                _throttledRequests.Remove(key);
```

waldekmastykarz (changes requested):

PR #252, opened 2023-04-25
Renames appsetings to mgdprc

waldekmastykarz:
Renames appsetings to mgdprc
Adds fallback to load mgdprc from appFolder


garrytrinder (approved):

gavinbarron:
I'm curious here, is this move to <exe-name>rc.json something that is becoming a convention in dotnet applications, or is there another motivation for this change?

waldekmastykarz:
Correct, we recommend that folks add the proxy installation folder to env path so that they can run it from any folder. But since we look for appsettings.json in the current working dir, to allow project-specific presets, if they'd run it from a working dir of another .net app, that would lead to collisions.

PR #253, opened 2023-04-25
Moves product version to code

waldekmastykarz:
Moves product version to code, to reliably detect new versions. Because .net doesn't stamp version numbers on assemblies on non-Windows OSes, we end up with the current version being 0.0.0.0.

### DIFF
--- a/msgraph-developer-proxy/ProxyEngine.cs
+++ b/msgraph-developer-proxy/ProxyEngine.cs
@@ -28,13 +28,10 @@ public class ProxyEngine {
     private static string _productVersion = string.Empty;
     public static string ProductVersion {
         get {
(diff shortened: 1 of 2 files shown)


garrytrinder (approved):

gavinbarron (approved):
I've been told that the version number should be getting stamped on all OSes and the behavior we're seeing on non-Windows builds is a bug.
That said I pivoted our build pipeline to used windows-latest a while back as a work around as I noticed this on Ubuntu based builds.
I think that this change is a good idea until we have an opportunity to re-visit this issue with the dotnet team.

waldekmastykarz:
Exactly! It would be awesome to be able to use the real product version as it's meant. Until then, we can use this workaround.

PR #257, opened 2023-05-02
Centralizes retry-after handling. Closes #239

### LINKED ISSUE #239: Re-using the Retry-After behavior across multiple plugins
We see multiple plugins doing the same (Rate Limiting, Chaos, etc.) with 429. We should have a better re-use of these capabilities.

waldekmastykarz:
Centralizes retry-after handling. Closes #239
Improves throttling


msgraph-developer-proxy-plugins/Behavior/RetryAfterPlugin.cs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    foreach (var throttler in e.ThrottledRequests)
+    {
+      var throttleInfo = throttler.ShouldThrottle(request, throttler.ThrottlingKey);
+      if (throttleInfo.ThrottleForSeconds > 0)
+      {
+        _logger?.LogRequest(new[] { $"Calling {request.Url} before waiting for the Retry-After period.", "Request will be throttled", $"Throttling on {throttler.ThrottlingKey}" }, MessageType.Failed, new LoggingContext(e.Session));
+
+        throttler.ResetTime = throttler.ResetTime.AddSeconds(throttleInfo.ThrottleForSeconds);
```
waldekmastykarz, line 61: ThrottleForSeconds needs to be set to DateTime.Now - ResetTime. Right now, we incorrectly manage expectations where we tell you to back off for 5 seconds while we're expecting you to wait reset time + 5 which leads to force fails all the time.

gavinbarron (approved):

PR #261, opened 2023-05-12
Adds detecting minimal permissions. Closes #57

### LINKED ISSUE #57: Help developers determine the minimal set of permissions for the (portion of the) app
When building apps that use Microsoft Graph, developers struggle with figuring out which set of permissions their app needs. They need to consider all the different Graph API calls that are issued on a page, for each one find out the necessary scopes, combine the scopes for the different calls and from there deduct the set of minimal permissions needed for the app. They need to do this for the different routes of their app, or for the whole app if they don't want to/can't use dynamic consent. This process is labor-intensive and error-prone.

The Proxy could simplify the process as follows:
- developers put the proxy in record mode, where the proxy keeps track of the different Graph API calls
- developers use their app to issue Graph API calls. Here, they can decide if they want to get scopes for the whole app or just a part of it
- developers stop the recording
- Proxy gets the list of unique endpoints called in the recorded session. For each endpoint, it retrieves the necessary permissions from the Graph dev API.
- Proxy makes a unique list of scopes necessary for the recorded endpoints and shows it to developers so that they can include it in their app

Depends on #93

waldekmastykarz:
Adds detecting minimal permissions. Closes #57

### DIFF
--- a/msgraph-developer-proxy-abstractions/ProxyUtils.cs
+++ b/msgraph-developer-proxy-abstractions/ProxyUtils.cs
@@ -1,15 +1,32 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Text.RegularExpressions;
 using Titanium.Web.Proxy.Http;
 using Titanium.Web.Proxy.Models;
 
 namespace Microsoft.Graph.DeveloperProxy.Abstractions;
 
+class ParsedSample {
+    public string QueryVersion { get; set; }
+    public string RequestUrl { get; set; }
+    public string SampleUrl { get; set; }
+    public string Search { get; set; }
+}
+
 public static class ProxyUtils {
-    public static bool IsGraphRequest(Request request) => 
-        request.RequestUri.Host.Contains("graph.microsoft.", StringComparison.OrdinalIgnoreCase) ||
-        request.RequestUri.Host.Contains("microsoftgraph.", StringComparison.OrdinalIgnoreCase);
+    private static readonly Regex itemPathRegex = new Regex(@"(?:\/)[\w]+:[\w\/.]+(:(?=\/)|$)");
+    private static readonly Regex sanitizedItemPathRegex = new Regex("^[a-z]+:<value>$", RegexOptions.IgnoreCase);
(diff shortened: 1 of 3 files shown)


sebastienlevert (changes requested):
My only request would be to utilize a kiota generated client to access the DevX API. Let's dogfood as much as possible!

waldekmastykarz:
> My only request would be to utilize a kiota generated client to access the DevX API. Let's dogfood as much as possible!

Do we have it already? If not, do we have all the metadata necessary to generate it, since the API is just in preview?

gavinbarron:
Hey @sebastienlevert while I like the idea of dogfooding kiota here, I'm not sure that this is the right project for it. 
It's very tempting from an engineering perspective to want to add this dependency, but that comes with costs:

* Engineering time to generate the client (if we can)
* Engineering time to integrate the additional dependencies including an anonymous auth provider for the request adapter
* Documentation time to make it easy for other contributors to work on this.
* Testing & Engineering cycles to validate the build/release process now that the plugins have additional dependencies that are not provided by the mgdp executable itself.
* Dependency churn increasing our dev effort every month as Kiota releases new versions consuming more engineering effort.

And all for one web request, one that we don't know if it has an Open API definition to generate from. It's a no from me at present.
I'd have a different opinion if this were for multiple APIs on the same domain or we had more time to devote to this project.

sebastienlevert:
We should have an OpenAPI for it. I understand your concerns, I still would love us to use Kiota here. These concerns are the concerns of any devs wanting to leverage an API. We are no different. If we vote that this is too much, then I think we failed with Kiota. Again, it's not a blocking issue, but we should be looking it at from the angle of let's validate our own patterns.

gavinbarron:
At the moment the current version of the OpenAPI definition is not correct for the permissions endpoint based on the shape of the returned object from a POST to permissions
Also a POST to the permissions endpoint on prod results in a 404 at present. 

Given that this is not in Prod yet we may need to hold this change until we have the change to the DevX API in Prod.

gavinbarron:
To move this forward, we're going to release a preview version of this plugin in the next release,.
We will address the migration to Kiota in #263 alongside the change to using the production version of the API in #264 when the DevX team release this change to production.

garrytrinder (approved):

PR #271, opened 2023-06-09
chore: capture initial branch policy settings

MIchaelMainer:
Capture current policy state so that we can make versioned changes to branch protection policies. This will enable policy history, auditing, review, standardization, and templatization. We will update policy in later PR.

We can ignore the rule that states: `At least one value under RequiredStatusChecks must be specified when RequiresStrictStatusChecks is true.` This is getting removed
* https://github.com/microsoft/GitOps.Branches/pull/73
* https://github.com/microsoft/GitOps.Branches/issues/72

This branch protection policy will not take effect until https://github.com/microsoftgraph/.github/blob/main/policies/branch-protection-sdks.yml has been removed.

### DIFF
--- a/.github/policies/msgraph-developer-proxy-branch-protection.yml
+++ b/.github/policies/msgraph-developer-proxy-branch-protection.yml
@@ -1,0 +1,42 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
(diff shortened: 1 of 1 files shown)


gavinbarron:
@MIchaelMainer perhaps we should close this as we have rehomed this repo into the microsoft org rather than in the microsoftgraph org?

PR #299, opened 2023-07-07
Adds support for batching in GraphRandomErrorPlugin. Closes #8

### LINKED ISSUE #8: Provide 424 failures for $batch requests
When a $batch request fails due to a throttling issue the request itself has a response status code of 424, while one or more requests within the batch have a status 429. When failing a $batch request the proxy should fail all requests within the batch with differing `Retry-After` values, the highest of these should be used to determine if the retried request may succeed.

Source: https://learn.microsoft.com/en-us/graph/throttling#throttling-and-batching

waldekmastykarz:
Adds support for batching in GraphRandomErrorPlugin. Closes #8


gavinbarron (approved):
This looks great!

PR #302, opened 2023-07-13
Extends select guidance with Open API info. Closes #107

### LINKED ISSUE #107: Show `$select` warning only for endpoints that support `$select`
> Feedback from Tom Morgan
> https://www.linkedin.com/feed/update/urn:li:activity:7018153776209371136/

Currently if a `GET` request passes through Microsoft Graph Developer Proxy that does not use the `$select` query string parameter we raise a warning in the proxy output.

This could lead to false positives as not all endpoints support `$select`.

waldekmastykarz:
Extends select guidance with Open API info. Closes #107

### DIFF
--- a/m365-developer-proxy-plugins/Guidance/GraphSelectGuidancePlugin.cs
+++ b/m365-developer-proxy-plugins/Guidance/GraphSelectGuidancePlugin.cs
@@ -1,37 +1,146 @@
 ﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+using System.Text.RegularExpressions;
 using Microsoft.Extensions.Configuration;
+using Microsoft.OpenApi.Models;
+using Microsoft.OpenApi.Readers;
 using Microsoft365.DeveloperProxy.Abstractions;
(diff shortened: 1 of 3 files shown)


m365-developer-proxy-plugins/Guidance/GraphSelectGuidancePlugin.cs
```diff
@@ -1,37 +1,146 @@
...
+        {
+            return false;
+        }
+
+        var graphVersion = ProxyUtils.GetGraphVersion(request.RequestUri.AbsoluteUri);
+        var tokenizedUrl = GetTokenizedUrl(request.RequestUri.AbsoluteUri);
+
+        if (EndpointSupportsSelect(graphVersion, tokenizedUrl))
+        {
+            return !request.Url.Contains("$select", StringComparison.OrdinalIgnoreCase) &&
```
gavinbarron, line 104: There's a missing case here, on beta `select` is also accepted without the `$` prefix

gavinbarron (approved):
Looks good to me.

PR #315, opened 2023-08-10
Renames plugins folder. Closes #314

### LINKED ISSUE #314: Rename plugins folder from GraphProxyPlugins to plugins
After we renamed the proxy to _Microsoft 365 Developer Proxy_ we forgot to include the name of the plugin folder, which is now named _GraphProxyPlugins_. Let's rename it to more generic `plugins` to avoid future renaming.

waldekmastykarz:
Renames plugins folder. Closes #314


garrytrinder:
I've also updated the folder references in the wiki.

garrytrinder (approved):

PR #317, opened 2023-08-10
Improve error handling when using options from disabled plugins. Closes #289

### LINKED ISSUE #289: Provide more accurate error message when using options from disabled plugins
We should also consider how we can provide a better message in scenarios like this.

`--summary-file-path` is a known option, so we should be able to provide a more helpful message, like

> To use `<option name>` you need to enable `<name of plugin>` plugin.

_Originally posted by @garrytrinder in https://github.com/microsoft/m365-developer-proxy/issues/282#issuecomment-1621492626_

garrytrinder:
Closes #289 

Merging this PR will

- Add error handling to compare the incoming short (-n) and long (--name) option aliases against the option aliases loaded from the proxy and plugins.
- No longer invoke `--help` output when an error is thrown.
- Add a TIP message to use `--help` to view available options.
- Add a TIP message with link to guidance for enabling plugins when a unknown option is used.

image

EDIT: 2023-08-10 15:47 - Addressed PR review comments
EDIT: 2023-08-11 09:08 - Addressed PR review comments

### DIFF
--- a/m365-developer-proxy/Program.cs
+++ b/m365-developer-proxy/Program.cs
@@ -10,6 +10,7 @@ ILogger logger = new ConsoleLogger(ProxyCommandHandler.Configuration, pluginEven
 IProxyContext context = new ProxyContext(logger, ProxyCommandHandler.Configuration);
 ProxyHost proxyHost = new();
 RootCommand rootCommand = proxyHost.GetRootCommand();
+
 PluginLoaderResult loaderResults = new PluginLoader(logger).LoadPlugins(pluginEvents, context);
 
 // have all the plugins init and provide any command line options
@@ -18,10 +19,11 @@ pluginEvents.RaiseInit(new InitArgs(rootCommand));
 rootCommand.Handler = proxyHost.GetCommandHandler(pluginEvents, loaderResults.UrlsToWatch, logger);
 
 // store the global options that are created automatically for us
-string[] globalOptions = { "--version", "-?", "-h", "--help" };
+// rootCommand doesn't return the global options, so we have to store them manually
+string[] globalOptions = { "--version", "--help", "-h", "/h", "-?", "/?" };
 
-// filter args to retrieve short (-n) and long (--name) option aliases
-var incomingOptions = args.Where(arg => arg.StartsWith("-") || arg.StartsWith("--")).ToArray();
+// filter args to retrieve options
+var incomingOptions = args.Where(arg => arg.StartsWith("-") || arg.StartsWith("/")).ToArray();
 
 // remove the global options from the incoming options
 incomingOptions = incomingOptions.Except(globalOptions).ToArray();
@@ -47,8 +49,8 @@ foreach (var option in rootCommand.Options)
 // list the remaining incoming options as unknown in the output
 if (incomingOptions.Length > 0)
 {
-    logger.LogInfo($"Unknown option(s): {string.Join(" ", incomingOptions)}");
-    logger.LogInfo($"TIP: Use -?, -h or --help to view available options");
+    logger.LogError($"Unknown option(s): {string.Join(" ", incomingOptions)}");
+    logger.LogInfo($"TIP: Use --help view available options");
     logger.LogInfo($"TIP: Are you missing a plugin? See: https://aka.ms/m365/proxy/plugins");
 }
 else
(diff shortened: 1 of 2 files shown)


m365-developer-proxy/Program.cs
```diff
@@ -17,4 +17,32 @@
 
 rootCommand.Handler = proxyHost.GetCommandHandler(pluginEvents, loaderResults.UrlsToWatch, logger);
 
-return await rootCommand.InvokeAsync(args);
\ No newline at end of file
+// filter args to retrieve short (-n) and long (--name) option aliases
+string[] incomingOptions = args.Where(arg => arg.StartsWith("-") || arg.StartsWith("--")).ToArray();
```
waldekmastykarz, line 21: We can simplify this with `var incomingOptions = ...`

m365-developer-proxy/Program.cs
```diff
@@ -17,4 +17,32 @@
...
+
+// compare the incoming options against the root command options
+foreach (Option option in rootCommand.Options)
```
waldekmastykarz, line 24: We can simplify this with `foreach (var option in ...`

waldekmastykarz (changes requested):
I like the idea, but let's clean it up before we merge:

image

- seems like we're showing the unmatched options twice
- `bar` is not an option but a value so we shouldn't include it in the list of unmatched options
- if we find an unmatched option, then perhaps we should exit early and not invoke the root command to avoid duplicate output

m365-developer-proxy/Program.cs
garrytrinder: Done ✅

waldekmastykarz (changes requested):
Almost, it breaks our help system:

image

We need to consider standard options that are a part of the underlying platform

m365-developer-proxy/Program.cs
```diff
@@ -17,4 +17,41 @@
 
 rootCommand.Handler = proxyHost.GetCommandHandler(pluginEvents, loaderResults.UrlsToWatch, logger);
 
-return await rootCommand.InvokeAsync(args);
\ No newline at end of file
+// store the global options that are created automatically for us
+string[] globalOptions = { "--version", "-?", "-h", "--help" };
```
waldekmastykarz, line 21: Seems like we're missing a few accepted variants: https://github.com/dotnet/command-line-api/blob/a045dd54a4c44723c215d992288160eb1401bb7f/src/System.CommandLine/Help/HelpOption.cs#L25

Let's add them

m365-developer-proxy/Program.cs
```diff
@@ -17,4 +17,41 @@
...
+
+// filter args to retrieve short (-n) and long (--name) option aliases
+var incomingOptions = args.Where(arg => arg.StartsWith("-") || arg.StartsWith("--")).ToArray();
```
waldekmastykarz, line 24: We can simplify this. An arg that starts with `--` also starts with `-` so the first check will already capture the second, which we can remove. Here, we should also consider the Windows-style switches with `/`

m365-developer-proxy/Program.cs
```diff
@@ -17,4 +17,41 @@
...
+            // remove the option from the incoming options
+            incomingOptions = incomingOptions.Where(val => val != alias).ToArray();
+        }
+    }
+}
+
+// list the remaining incoming options as unknown in the output
+if (incomingOptions.Length > 0)
+{
+    logger.LogInfo($"Unknown option(s): {string.Join(" ", incomingOptions)}");
```
waldekmastykarz, line 50: These statements seem more like messages than actual command output. As such, I think we should output them to stderr rather than stdout. We can do it by using LogError instead. It will mean though that they'd be colored red. It could be useful to bring the attention to it (after all, something went wrong), but if it's too excessive, then we should consider introducing an extra logging method on `ILogger` which uses default printing style but sends the output to stderr.

garrytrinder:
Thanks again for the testing and feedback @waldekmastykarz.

I've updated the PR description to reflect the changes made.

m365-developer-proxy/Program.cs
garrytrinder: Done

m365-developer-proxy/Program.cs
garrytrinder: Done

m365-developer-proxy/Program.cs
garrytrinder: Done.

garrytrinder:
I've also created the page behind the aka.ms short link.

waldekmastykarz (approved):

PR #325, opened 2023-08-28
Adds support for blocking unmocked requests. Closes #311

### LINKED ISSUE #311: Extend mock plugins with config to block passing thru unmocked requests
Extend mock plugins with config to block passing thru unmocked requests. By default, unmocked requests are passed thru, but when passing thru is disabled, the mock plugin will return a `502 Bad Gateway` response.

### Background

Should we have a mode where proxy ONLY mocks and never calls Graph? Right now AFAIK we are providing an hybrid where "if" there is a mock, we return the mock but if not, we return Graph data right? This behavior would be different from the one suggested, though...

_Originally posted by @sebastienlevert in https://github.com/microsoft/m365-developer-proxy/issues/307#issuecomment-1671891857_

waldekmastykarz:
Adds support for blocking unmocked requests. Closes #311


waldekmastykarz:

gavinbarron (approved):

PR #326, opened 2023-08-28
Adds mocks generator plugin. Closes #305

### LINKED ISSUE #305: Create mocks from intercepted requests
Create a new summary plugin that creates mocks from the intercepted requests. This will allow users to create mocks based on their data.

After generating the mocks, the plugin sorts them descending by URL length to put the most specific URLs upfront and more generic URLs in the back. This will allow users to replace IDs with wildcards to generalize their mocks and ensure that they will execute in the correct order.

waldekmastykarz:
Adds mocks generator plugin. Closes #305

### DIFF
--- a/m365-developer-proxy-plugins/MockResponses/MockResponsePlugin.cs
+++ b/m365-developer-proxy-plugins/MockResponses/MockResponsePlugin.cs
@@ -16,7 +16,9 @@ using Titanium.Web.Proxy.Models;
 namespace Microsoft365.DeveloperProxy.Plugins.MocksResponses;
 
 internal class MockResponseConfiguration {
+    [JsonIgnore]
     public bool NoMocks { get; set; } = false;
+    [JsonIgnore]
     public string MocksFile { get; set; } = "responses.json";
 
     [JsonPropertyName("responses")]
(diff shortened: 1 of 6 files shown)


waldekmastykarz:

m365-developer-proxy-plugins/MockResponses/MockResponsePlugin.cs
```diff
@@ -16,7 +16,9 @@
 namespace Microsoft365.DeveloperProxy.Plugins.MocksResponses;
 
 internal class MockResponseConfiguration {
+    [JsonIgnore]
     public bool NoMocks { get; set; } = false;
+    [JsonIgnore]
```
gavinbarron, line 21: won't this prevent passing the location of the mock file via config?

waldekmastykarz: The way I understood it, is that it's for serialization, but let me double check that it doesn't affect deserialization. Good catch

waldekmastykarz: Just verified and it doesn't prevent configuring the mocks file via config

gavinbarron (approved):
LGTM

PR #328, opened 2023-08-28
Adds the GraphMockResponsePlugin. Closes #307

### LINKED ISSUE #307: Add support for mocking Graph batch responses
Extend the proxy with support for mocking responses inside Graph batch requests.

Right now, we support mocking requests based on their URL and method. For Graph batch requests this is insufficient, because while the method and the URL are always the same, the body of the request actually describes which endpoints are being called.

We should create a new plugin capable of mocking Graph batch responses. It would use the existing mock infrastructure, and on top add the ability to detect batch requests, unpack their requests, mock responses, build a batch response and return it to the client.

Before we proceed with implementation, we need to decide how to handle requests for which we have no mocks. We could do it in a few ways:

1. Pass the batch request through to Graph, intercept the response and replace what we can with mocks
2. Mock the requests that we can. Manually execute requests to Graph that we can't mock and stitch everything together into a mock batch response
3. Mock the requests that we can and return a 404 response for requests that we can't mock

Approach no. 3 is the easiest to build. It's the most accurate but it gives the user sufficient information to see which mocks are missing.
Approaches 1 and 2 give the user a more accurate response but are significantly more complex and error prone to implement.

I suggest that we started with no. 3.

waldekmastykarz:
Adds the GraphMockResponsePlugin. Closes #307

### DIFF
--- a/m365-developer-proxy-abstractions/GraphBatchResponsePayload.cs
+++ b/m365-developer-proxy-abstractions/GraphBatchResponsePayload.cs
@@ -16,7 +16,7 @@ public class GraphBatchResponsePayloadResponse {
     [JsonPropertyName("status")]
     public int Status { get; set; } = 200;
     [JsonPropertyName("body")]
-    public GraphBatchResponsePayloadResponseBody? Body { get; set; }
+    public dynamic? Body { get; set; }
     [JsonPropertyName("headers")]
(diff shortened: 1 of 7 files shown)


waldekmastykarz:

gavinbarron (approved):
LGTM

PR #353, opened 2023-09-21
Add RateLimitingPlugin config section

garrytrinder:
Merging this PR will

- Add missing `configSection` for `RateLimitingPlugin`


m365-developer-proxy/m365proxyrc.json
```diff
@@ -153,6 +153,11 @@
     "minMs": 200,
     "maxMs": 10000
   },
+  "rateLimiting": {
+    "costPerRequest": 1,
+    "rateLimit": 40,
```
waldekmastykarz, line 158: Let's align these values with defaults in the plugin

waldekmastykarz (changes requested):

m365-developer-proxy/m365proxyrc.json
garrytrinder: Good idea. I took this config from the example in plugin page in the docs so I'll update that as well 👍

garrytrinder: @waldekmastykarz updated 👍

waldekmastykarz (approved):

PR #357, opened 2023-10-02
chore: No more painfully to see warnings

svrooij:
It's just that ☝️ I cleared up the build warnings so your [github actions result](https://github.com/microsoft/m365-developer-proxy/actions/runs/6335359687#annotations) will no longer look like:

image


waldekmastykarz:
Thank you! We'll have a look asap

m365-developer-proxy-plugins/m365-developer-proxy-plugins.csproj
```diff
@@ -7,6 +7,7 @@
     <EnableDynamicLoading>true</EnableDynamicLoading>
     <RunPostBuildEvent>OnOutputUpdated</RunPostBuildEvent>
     <Version>0.12.0</Version>
+    <NoWarn>CS1998</NoWarn>
```
gavinbarron, line 10: Personally I'd rather have explicit pragma comments disabling specific warnings on this one rather than potentially hiding a bunch cases where there should be `await` or the method should not use async.

m365-developer-proxy-abstractions/PluginEvents.cs
```diff
@@ -187,15 +187,21 @@ public class PluginEvents : IPluginEvents {
     }
 
     public async Task RaiseProxyBeforeRequest(ProxyRequestArgs args) {
-        await BeforeRequest?.InvokeAsync(this, args, null);
+        if (BeforeRequest is not null) {
```
gavinbarron, line 190: Thanks for cleaning these up! It's much appreciated

m365-developer-proxy-plugins/m365-developer-proxy-plugins.csproj
svrooij: I added this one, because the plugins project if full with this one. I thought one row in the project vs several line ignored on all the occasions.

How about adding the Release condition? Then you still have the warnings during development and they won't show up in the build pipeline.

m365-developer-proxy-abstractions/PluginEvents.cs
svrooij: I had a discussion on this topic on X the other day, and we actually prefer the old way, but that causes a warning.

m365-developer-proxy-plugins/m365-developer-proxy-plugins.csproj
gavinbarron: That's better. I've just seen these kind of blanket suppressions for a project cause more problems than having the warnings visible to be comfortable with suppressing them entirely

gavinbarron:
```suggestion
    <PropertyGroup Condition="'$(Configuration)'=='Release'">
      <NoWarn>CS1998</NoWarn>
    </PropertyGroup>
```

gavinbarron: Actually, it needs a separate PropertyGroup with the condition, nesting like the suggestion doesn't work

svrooij:
@waldekmastykarz I got an update that you merged the latest main to this branch, but there aren’t any updates right?

waldekmastykarz:
I got a message that there were updates on main so added them. I'll have a closer look at the PR asap 👍

waldekmastykarz (approved):

Issue #301, opened 2023-07-13
Show post body


kiwinol:
I would like to be able to see the body content for post requests.

Would this be possible?

waldekmastykarz:
Just to check if I understand it correctly, when the proxy intercepts a request, you'd like to be able to see the body of the request that's going over the wire, correct? Would you also like to see the headers or just the body? What would you like to do with this information?

kiwinol:
Yes, I would like to see the body when the proxy intercepts a request over the wire.

For my use case I think headers are not required just the body.

I am using PowerShell and Invoke-RestMethod with a post and body to a reporting endpoint this is working.
A colleague is using C# and the official Microsoft SDK.  He believes posting the same body but always receives 0 results.

I would like to compare the Body I am sending via PowerShell with the body that my colleague is sending via the C# library.

Hopefully this makes sense :)

garrytrinder:
Thanks for the further context @kiwinol but I wonder how readable this would be in the standard output.

My initial thoughts are to add a new option to configure the summary report output to include the body.

Any thoughts or ideas on this? @sebastienlevert @gavinbarron

waldekmastykarz:
It does @kiwinol, thank you for explaining. @garrytrinder, considering @kiwinol scenario and your suggestion of using the summary report, you'd need to do the following:

- start proxy
- start recording
- execute the request
- stop recording
- open the report

In comparison, if we were to show the body directly in the output, it would be:

- start proxy
- execute the request

I fully agree that we'd need to consider readability and how to make it work in a terminal which doesn't have anchors and any other UI to indicate the beginning and end of each request. Let's look at possible solutions before we commit to a direction.

garrytrinder:
> Let's look at possible solutions before we commit to a direction.

Absolutely.

If we were to consider adding the output to the terminal then it seems like exposing a new option `--showBody` and/or a configurable setting in `m365proxyrc.json` would be good candidates.

We could also enhance the `ExecutionSummaryPlugin` to respect this setting to return the body in the summary report should this be needed.

waldekmastykarz:
I've been thinking about this some more, and would like to suggest that we use the [Chrome DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/) for this. It gives us a UI that's de facto market standard and allows us to present the necessary information without cluttering the terminal output. I'll see if I can create a POC to better understand how it would work.

PR #365, opened 2023-10-23
Adds inspection with CDP

waldekmastykarz:
Related to #301

### DIFF
--- a/m365-developer-proxy-plugins/Inspection/DevToolsPlugin.cs
+++ b/m365-developer-proxy-plugins/Inspection/DevToolsPlugin.cs
@@ -117,14 +117,19 @@ public class DevToolsPlugin : BaseProxyPlugin
             if (message?.Method == "Network.getResponseBody")
             {
                 var requestId = message.Params?.RequestId;
-                if (requestId is null || !responseBody.ContainsKey(requestId))
+                if (requestId is null ||
+                    !responseBody.ContainsKey(requestId) ||
+                    // should never happen because the message is sent from devtools
+                    // and Id is required on all socket messages but theoretically
+                    // it is possible
+                    message.Id is null)
                 {
                     return;
                 }
 
                 var result = new GetResponseBodyResult
                 {
-                    Id = (int)message.Id!,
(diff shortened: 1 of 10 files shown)


m365-developer-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -0,0 +1,299 @@
...
+            {
+                var requestId = message.Params?.RequestId;
+                if (requestId is null || !responseBody.ContainsKey(requestId))
+                {
+                    return;
+                }
+
+                var result = new GetResponseBodyResult
+                {
+                    Id = (int)message.Id!,
```
gavinbarron, line 127: Should we be checking for the Id on the message in the early exit conditions on 120? or providing some form of logical default?

m365-developer-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        var message = new EntryAddedMessage
+        {
+            Params = new()
+            {
+                Entry = new()
+                {
+                    Source = "network",
+                    Text = string.Join(" ", e.RequestLog.Message),
```
gavinbarron, line 289: *nit:* perhaps we should rename this `Message` property to be `Messages` or `MessageLines`?

waldekmastykarz: MessageLines would be clearer. Agreed

m365-developer-proxy-plugins/Inspection/DevToolsPlugin.cs
waldekmastykarz: We could do early exit but it's a part of the spec that a message contains an ID which is why we're assuming it here. If dev tools would send us a message without it, they'd be breaking spec and we wouldn't be able to answer it because messages are matched on ID.

gavinbarron (approved):
👍

PR #382, opened 2023-11-13
Renames to Dev Proxy

waldekmastykarz:


gavinbarron (approved):
LGTM, just need to ensure that we have the code owners group renamed to match our new names

PR #388, opened 2023-11-16
Adds IP Address option. Closes #380

### LINKED ISSUE #380: [BUG]: error: "IPv4 address 0.0.0.0 and IPv6 address ::0 are unspecified addresses that cannot be used as a target address. (Parameter 'hostName') (0.0.0.0:8000)"
### Description

When you register the proxy as a system proxy, you get at some point the following error:

```text
IPv4 address 0.0.0.0 and IPv6 address ::0 are unspecified addresses that cannot be used as a target address. (Parameter 'hostName') (0.0.0.0:8000)
IPv4 address 0.0.0.0 and IPv6 address ::0 are unspecified addresses that cannot be used as a target address. (Parameter 'hostName') (0.0.0.0:8000)
```

### Expected behaviour

No error

### Actual behaviour

Bug

### Steps to reproduce

1. Configure proxy as a system proxy (extra step on macOS, default behavior on Windows)
2. Wait until the error

### Microsoft 365 Developer Proxy Version

latest from main

### Operating system (environment)

macOS

### Shell

zsh

### Additional Info

The error stems from `m365-developer-proxy/MSGraphDbCommandHandler.cs:166`. It seems like our HTTP client responsible for updating Graph Open API files can't handle the proxy correctly and throws this error. While the proxy works just fine, updating the local DB fails.

waldekmastykarz:
Adds IP Address option. Closes #380

### DIFF
--- a/dev-proxy-abstractions/IProxyConfiguration.cs
+++ b/dev-proxy-abstractions/IProxyConfiguration.cs
@@ -16,6 +16,7 @@ public enum LabelMode {
 
 public interface IProxyConfiguration {
     int Port { get; }
+    string? IPAddress { get; }
     LabelMode LabelMode { get; }
     bool Record { get; }
     LogLevel LogLevel { get; }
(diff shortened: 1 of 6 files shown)


waldekmastykarz:
@garrytrinder we also should add to our docs, that on macOS, folks should configure proxy to listen on a specific port rather than 0.0.0.0, because it breaks updating the Graph DB and looking up minimal permissions.

PR #413, opened 2023-12-08
Updates logLevel to include plugins. Closes #167

### LINKED ISSUE #167: `--logLevel debug` doesn't show plugin loading information
Right now, when you use `--logLevel debug` when starting the proxy, you won't see debug information from loading plugins. This is because plugins are loaded before we process options. This is confusing for our users because since we expose the option, it gives the impression that the option is an full equivalent of the setting in appsettings.json.

@gavinbarron could we solve this by processing options twice: once before we do anything else, and then second time after we loaded plugins?

waldekmastykarz:
Updates logLevel to include plugins. Closes #167


garrytrinder (approved):
Plugins shown in output using `--log-level debug`

image

Friendly error shown when using an invalid log level.

image

PR #430, opened 2023-12-15
Adds setup scripts

waldekmastykarz:
Adds setup scripts for bash (macOS and Linux) and PowerShell (Windows, macOS and Linux)

To test, run:

```sh
curl -sL https://raw.githubusercontent.com/waldekmastykarz/dev-proxy/setup-script/scripts/setup.sh | bash
```

or

```powershell
(invoke-webrequest https://raw.githubusercontent.com/waldekmastykarz/dev-proxy/setup-script/scripts/setup.ps1).Content | Invoke-Expression
```


garrytrinder (changes requested):
I tested the PS1 setup script, but dot sourcing my profile does not enable my to run proxy from any folder.

image

image

I can run proxy fine from the installation directory however.

I also noticed that we automatically change directory for the user to the `devproxy` directory, I wonder if we should do that and instead keep them in their current directory and provide the location to the `devproxy` installation files in the output.

Tested on Windows 11 Insider Preview and PowerShell 7.4.0

scripts/setup.ps1
```diff
@@ -0,0 +1,71 @@
...
+Add-Type -AssemblyName System.IO.Compression.FileSystem
+Expand-Archive -Path devproxy.zip -DestinationPath . -Force -ErrorAction Stop
+Remove-Item devproxy.zip
+
+if (!(Test-Path $PROFILE)) {
+    New-Item -ItemType File -Force -Path $PROFILE
+}
+
+if (!(Select-String -Path $PROFILE -Pattern "devproxy")) {
+    Add-Content -Path $PROFILE -Value "`$env:PATH += `":$full_path`""
```
garrytrinder, line 64: This results in a broken PATH string. 

We should remove the `:` from start and replace with a following `;`

```suggestion
    Add-Content -Path $PROFILE -Value "`$env:PATH += `"$full_path;`""
```

garrytrinder (commented):
Tested the bash script, that works well.

As it's possible for mac users to use either the bash or PS1 script, we should consider outputting the below instructions to update the network adaptor, or provide a short link to a page in docs with these instructions documented.

> Configure your network device.
> Open Network settings.
> Select the device to configure and select the Advanced... button.
> Go to the Proxies tab.
> Check Secure Web Proxy (HTTPS) in the list of configurable proxies.
> Enter 127.0.0.1:8000 in the Secure Web Proxy Server field.
> Select OK and then Apply to confirm the changes.

scripts/setup.ps1
waldekmastykarz: Are you sure? In my experience, at least on macOS, you need to use `:` instead of `;`. `;` on macOS leads to errors

garrytrinder (commented):
The scripts are working fine, but I think we should reconsider some of the locations used and behaviours. These are all non-blocking comments.

- Rather than downloading the ZIP file to the current directory, we should consider using the users local temp directory in which to download the ZIP file to. 
- Rather than creating the `devproxy` folder in the current directory, we should consider creating the `devproxy` folder in the users home directory as a default location.
- Rather than changing into the directory to unzip the file, we should consider using a full path with the `Expand-Archive` cmdlet, which means that we don't have to change the directory on the users behalf. The Windows script changes the user directory, whereas the bash script doesn't.
- We should consider outputting the full path the `devproxy` folder so that users know what this script has just done on their behalf. It's not obvious where files have been installed to from the current console output.

garrytrinder (approved):

PR #431, opened 2023-12-16
Adds cert setup for macOS. Closes #49

### LINKED ISSUE #49: Simplify macOS setup
Enabling the proxy on macOS requires manual steps. In comparison, on Windows, after starting the proxy, it automatically registers as a system-wide proxy. We should look for ways to simplify the steps, either via a script people can run, or have the proxy automatically register as a system-wide proxy, just like it does on Windows.

waldekmastykarz:
Adds cert setup for macOS. Closes #49

- adds a bash script that Dev Proxy runs the first time it's started to configure its certificate as trusted on macOS (if needed, users can run the script in the future manually, or reset the first-run state by deleting the `.hasrun` file in Dev Proxy app folder
- adds `--no-first-run` option to suppress first-run experience in cases Dev Proxy can't run interactively
- renames Dev Proxy certificate to `Dev Proxy CA`


dev-proxy/trust-cert.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo -e "\n"
+
+cert_name="Dev Proxy CA"
+# export cert from keychain to PEM
+echo "Exporting Dev Proxy certificate..."
+security find-certificate -c "$cert_name" -a -p > dev-proxy-ca.pem
+# add trusted cert to keychain
+echo "Updating Dev Proxy trust settings..."
+security add-trusted-cert -r trustRoot -k ~/Library/Keychains/login.keychain-db dev-proxy-ca.pem
```
garrytrinder, line 19: This line fails with the following error...

```
SecCertificateCreateFromData: One or more parameters passed to a function were not valid.
Error reading file dev-proxy-ca.pem
```

`dev-proxy-ca.pem` is in the working directory.

I get the same error if I run the `security add-trusted-cert` command directly in the terminal. I also get the same error when using `sudo`.

dev-proxy/trust-cert.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo -e "\n\033[1;33mTrust the certificate in your Keychain manually to avoid errors.\033[0m\n"
+    exit 1
+fi
+
+echo -e "\n"
+
+cert_name="Dev Proxy CA"
+# export cert from keychain to PEM
+echo "Exporting Dev Proxy certificate..."
+security find-certificate -c "$cert_name" -a -p > dev-proxy-ca.pem
```
garrytrinder, line 16: This generates a zero byte file.

dev-proxy/trust-cert.sh
```diff
@@ -0,0 +1,23 @@
...
+read -n 1 answer
+
+if [ "$answer" = "n" ]; then
+    echo -e "\n\033[1;33mTrust the certificate in your Keychain manually to avoid errors.\033[0m\n"
+    exit 1
+fi
+
+echo -e "\n"
+
+cert_name="Dev Proxy CA"
```
garrytrinder, line 13:
```suggestion
cert_name="Titanium Root Certificate Authority"
```

garrytrinder (changes requested):

garrytrinder (approved):

PR #443, opened 2023-12-20
Update default config

garrytrinder:
- Change default config to placeholder API with generic random errors
- Move m365 config to a separate preset


dev-proxy/dev-proxy.csproj
```diff
@@ -50,27 +50,32 @@
...
+    <None Update="presets\microsoft-365.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Update="presets\microsoft-graph-rate-limiting.json">
+    <None Update="presets\microsoft-365-errors.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
 
+  <ProjectExtensions><VisualStudio><UserProperties default_1json__JsonSchema="https://json.schemastore.org/global.json" devproxyrc_1json__JsonSchema="https://json.schemastore.org/global.json" errors_1json__JsonSchema="https://json.schemastore.org/global.json" /></VisualStudio></ProjectExtensions>
```
waldekmastykarz, line 79: Do we need this?

dev-proxy/devproxyrc.json
```diff
@@ -2,198 +2,22 @@
...
-  },
-  "devTools": {
-    "preferredBrowser": "Edge"
-  },
-  "rateLimiting": {
-    "costPerRequest": 2,
-    "rateLimit": 120,
-    "retryAfterSeconds": 5
-  },
-  "rate": 50,
```
waldekmastykarz, line 196: Don't we want to keep the global settings for easy change?

dev-proxy/presets/m365.json
```diff
@@ -0,0 +1,195 @@
+{
+    "$schema": "https://raw.githubusercontent.com/microsoft/dev-proxy/main/schemas/v1.0/rc.schema.json",
+    "plugins": [
+      {
+        "name": "DevToolsPlugin",
+        "enabled": false,
+        "pluginPath": "plugins\\dev-proxy-plugins.dll",
```
waldekmastykarz, line 7: I think this will break when you try to use it from a folder other than where proxy is installed. Let's prefix it with `~appFolder\\` so that proxy always resolves it to its installation folder and calling this preset will work from any folder.

waldekmastykarz (changes requested):
Nice! Let's do a few small adjustments

dev-proxy/dev-proxy.csproj
garrytrinder: No, this snuck in, good catch. I'll remove.

dev-proxy/devproxyrc.json
garrytrinder: Good catch, I'll add `rate` and also `labelMode` and `logLevel`.

dev-proxy/presets/m365.json
garrytrinder: Good catch, I'll update the paths.

waldekmastykarz (approved):

PR #503, opened 2024-01-18
Align response objects in schemas

garrytrinder:
Merging this PR will align the response objects used in both `mockresponseplugn.schema.json ` and `genericrandomerrorplugin.schema.json`

### DIFF
--- a/schemas/v1.0/mockresponseplugin.schema.json
+++ b/schemas/v1.0/mockresponseplugin.schema.json
@@ -70,11 +70,7 @@
                   ]
                 }
               }
-            },
-            "required": [
-              "statusCode",
-              "body"
-            ]
+            }
           }
         },
         "required": [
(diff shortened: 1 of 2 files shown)


schemas/v1.0/mockresponseplugin.schema.json
```diff
@@ -41,33 +41,40 @@
...
+                  },
+                  "required": [
+                    "name",
+                    "value"
+                  ]
+                }
               }
-            }
+            },
+            "required": [
```
waldekmastykarz, line 74: None of the properties are required. In theory, all of them could be empty and you'd get a 200 OK response. So let's remove it.

waldekmastykarz (changes requested):

waldekmastykarz (approved):

PR #511, opened 2024-01-22
Adds permissions to address GH change

waldekmastykarz:
Adds permissions to address this breaking change in GH: https://docs.opensource.microsoft.com/github/apps/permission-changes/.

Let's validate if we've got everything. As far as I can tell, we only need write permission for contents to create a release, but let's double check.


gavinbarron:
Yep, release is the only case where we need a write token in this repo at present.

gavinbarron (approved):

Issue #527, opened 2024-01-26
When plugin defines an empty array of URLs to watch, use the global one


waldekmastykarz:
Right now, we allow for plugins to define an empty array of URLs to watch: https://github.com/microsoft/dev-proxy/blob/e5471ecbe1f70dfae554d7c5f8062131cede767a/dev-proxy/PluginLoader.cs#L48. On runtime, it'll be equivalent of a plugin being disabled because there won't be a URL to watch. This is confusing and undesirable. We should change the code so that it doesn't just check for null but also for empty array, and in both cases defaults to the global list of URLs to watch.

PR #528, opened 2024-01-26
When plugin defines an empty array of URLs to watch, use the global o…

SilentSobs:
…ne #527

Checking for Null or Empty:

The first if statement checks if pluginUrlsList is either null or empty.
If it's null or empty, it sets pluginUrlsList to the global list of URLs (globalUrlsToWatch.ToList()).
Processing Non-Empty List:

The second if statement checks if pluginUrlsList has any elements.
If it has elements, it processes the non-empty pluginUrlsList.
This modification ensures that even if the plugin defines an empty array of URLs, the code will not consider the plugin "disabled." Instead, it defaults to using the global list of URLs, preventing confusion and ensuring that there is always a valid list for further processing.

### DIFF
--- a/dev-proxy/PluginLoader.cs
+++ b/dev-proxy/PluginLoader.cs
@@ -52,7 +52,12 @@ internal class PluginLoader
                 }
                 // Load Plugins from assembly
                 IProxyPlugin plugin = CreatePlugin(assembly, h);
-                plugin.Register(pluginEvents,proxyContext,pluginUrls is not null && pluginUrls.Any() ? pluginUrls : defaultUrlsToWatch,h.ConfigSection is null ? null : Configuration.GetSection(h.ConfigSection));
+                plugin.Register(
+                    pluginEvents,
+                    proxyContext,
+                    (pluginUrls != null && pluginUrls.Count > 0) ? pluginUrls : defaultUrlsToWatch,
+                    h.ConfigSection is null ? null : Configuration.GetSection(h.ConfigSection)
+                );
                 plugins.Add(plugin);
             }
         }
--- a/dev-proxy/PluginLoader.cs
+++ b/dev-proxy/PluginLoader.cs
@@ -45,8 +45,11 @@ internal class PluginLoader
                 Assembly assembly = pluginLoadContext.LoadFromAssemblyName(new AssemblyName(Path.GetFileNameWithoutExtension(pluginLocation)));
                 IEnumerable<UrlToWatch>? pluginUrlsList = h.UrlsToWatch?.Select(ConvertToRegex);
                 ISet<UrlToWatch>? pluginUrls = null;
-                if (pluginUrlsList is not null)
-                {
+                if (pluginUrlsList is null || !pluginUrlsList.Any()){
+                    pluginUrlsList = globalUrlsToWatch.ToList();
+                }
+
+                if (pluginUrlsList.Any()){
                     pluginUrls = pluginUrlsList.ToHashSet();
                     globallyWatchedUrls.AddRange(pluginUrlsList);
                 }
--- a/dev-proxy/PluginLoader.cs
+++ b/dev-proxy/PluginLoader.cs
@@ -45,17 +45,14 @@ internal class PluginLoader
                 Assembly assembly = pluginLoadContext.LoadFromAssemblyName(new AssemblyName(Path.GetFileNameWithoutExtension(pluginLocation)));
                 IEnumerable<UrlToWatch>? pluginUrlsList = h.UrlsToWatch?.Select(ConvertToRegex);
                 ISet<UrlToWatch>? pluginUrls = null;
-                if (pluginUrlsList is null || !pluginUrlsList.Any()){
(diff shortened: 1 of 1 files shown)


dev-proxy/PluginLoader.cs
```diff
@@ -45,8 +45,11 @@ public PluginLoaderResult LoadPlugins(IPluginEvents pluginEvents, IProxyContext
                 Assembly assembly = pluginLoadContext.LoadFromAssemblyName(new AssemblyName(Path.GetFileNameWithoutExtension(pluginLocation)));
                 IEnumerable<UrlToWatch>? pluginUrlsList = h.UrlsToWatch?.Select(ConvertToRegex);
                 ISet<UrlToWatch>? pluginUrls = null;
-                if (pluginUrlsList is not null)
-                {
+                if (pluginUrlsList is null || !pluginUrlsList.Any()){
```
waldekmastykarz, line 48: Perfect! Let's move { to the next line for consistent formatting and we'll get it merged! We appreciate your help!

waldekmastykarz (changes requested):

dev-proxy/PluginLoader.cs
```diff
@@ -45,8 +45,11 @@ public PluginLoaderResult LoadPlugins(IPluginEvents pluginEvents, IProxyContext
...
+                    pluginUrlsList = globalUrlsToWatch.ToList();
+                }
+
+                if (pluginUrlsList.Any()){
```
waldekmastykarz, line 52: Since we've checked the pluginUrlList already above, why checking it again here? When would this check be falsy?

Also, let's drop the { to the next line for consistency.

dev-proxy/PluginLoader.cs
waldekmastykarz, line 48: Rather than changing the logic, it would be easier to add `&& pluginUrlList.Any()` to the original code which would meet the requirement.

waldekmastykarz (changes requested):

SilentSobs:
```


List<UrlToWatch> globallyWatchedUrls = new List<UrlToWatch>();

if (pluginUrlsList is not null && !pluginUrlsList.Any())
{
    // Set pluginUrlsList to globallyWatchedUrls.ToList() if it's null or empty
    pluginUrlsList = globallyWatchedUrls.ToList();
}
else if (pluginUrlsList?.Any() == true)
{
    // Directly check if pluginUrlsList is not null and contains elements
    pluginUrls = pluginUrlsList.ToHashSet();
    globallyWatchedUrls.AddRange(pluginUrlsList);
}

```

dev-proxy/PluginLoader.cs
SilentSobs: > Perfect! Let's move { to the next line for consistent formatting and we'll get it merged! We appreciate your help!

```
if (pluginUrlsList is not null && !pluginUrlsList.Any())
        {
            // Set pluginUrlsList to globallyWatchedUrls.ToList() if it's null or empty
            pluginUrlsList = globallyWatchedUrls.ToList();
        }
        else if (pluginUrlsList?.Any() == true)
        {
            // Directly check if pluginUrlsList is not null and contains elements
            pluginUrls = pluginUrlsList.ToHashSet();
            globallyWatchedUrls.AddRange(pluginUrlsList);
        }
```

SilentSobs:
Please Check @waldekmastykarz 

if (pluginUrlsList is not null && !pluginUrlsList.Any())
{
    pluginUrlsList = globallyWatchedUrls.ToList();
}
else if (pluginUrlsList?.Any() == true)
{

    pluginUrls = pluginUrlsList.ToHashSet();
    globallyWatchedUrls.AddRange(pluginUrlsList);
}

SilentSobs:
@waldekmastykarz New To This stuff sorry please review this once 

```
if (pluginUrlsList is not null && !pluginUrlsList.Any())
{

    pluginUrlsList = globallyWatchedUrls;
}
else if (pluginUrlsList?.Any() == true)
{
    pluginUrls = pluginUrlsList.ToHashSet();
    globallyWatchedUrls.AddRange(pluginUrlsList);
}
```

waldekmastykarz:
Sorry, my bad. Adding an empty array of URLs from the plugin to the list of global URLs is not an issue so we can just leave it be as it originally was.

What we need to change is how we pass URLs to plugin registration on line 55:

pluginUrls ?? defaultUrlsToWatch

Needs to change to include checking not only for pluginsUrl being null but also not empty.

Sorry for the confusion.

SilentSobs:
> Sorry, my bad. Adding an empty array of URLs from the plugin to the list of global URLs is not an issue so we can just leave it be as it originally was.
> 
> What we need to change is how we pass URLs to plugin registration on line 55:
> 
> pluginUrls ?? defaultUrlsToWatch
> 
> Needs to change to include checking not only for pluginsUrl being null but also not empty.
> 
> Sorry for the confusion.

Pull request submitted (#528). Please review at your convenience. Thank you!

PR #530, opened 2024-01-26
Add option to not run as system proxy and do not install self-signed cert

ohaucke:
Adds the following two options:
* `--do-not-act-as-system-proxy` - `Do not set the proxy as system proxy`
* `--do-not-install-self-signed-cert` - `Do not install self-signed certificate`

The option `--do-not-install-self-signed-cert` depends on `--do-not-act-as-system-proxy` and will return the following error message when it's called without it:
`Requires option '--do-not-act-as-system-proxy' to be 'true'`

That's due to `_proxyServer.SetAsSystemHttpsProxy(_explicitEndPoint)` which internally calls `CertificateManager.EnsureRootCertificate()` when `ProxyProtocolType.Https` is set.


ohaucke:
@microsoft-github-policy-service agree

dev-proxy/ProxyCommandHandler.cs
```diff
@@ -18,6 +18,8 @@ public class ProxyCommandHandler : ICommandHandler
     public Option<IEnumerable<string>?> WatchProcessNames { get; set; }
     public Option<int?> Rate { get; set; }
     public Option<bool?> NoFirstRun { get; set; }
+    public Option<bool?> DoNotActAsSystemProxy { get; set; }
```
waldekmastykarz, line 21: Let's rename these options to be "positive". Otherwise, we end up with double negations (`!DoNotAct...`) which makes it hard to follow the code. So instead, we'd have `ActAsSystemProxy` with a default value of `true` and you could set it to `false` if you needed

waldekmastykarz (changes requested):

waldekmastykarz:
Nice start! Let's see how the code changes after we rename options to be "positive".

dev-proxy/ProxyCommandHandler.cs
waldekmastykarz: Yes, I find this clearer. Also, I wonder if there are some other, less verbose names for these options, that are just as clear. Would `--system-proxy false` and `--install-cert false` be just as clear?

ohaucke: How about `--as-system-proxy` and `--install-cert`?
`--system-proxy` could be misunderstood as "do not use the currently configured system proxy".

waldekmastykarz: Good rationale. Let's use `--as-system-proxy` and `--install-cert`

ohaucke: Changed with 75ff1107c3ccd13948cc18d4f9a661d3fd1f2428

dev-proxy/ProxyEngine.cs
```diff
@@ -95,19 +99,14 @@ public async Task Run(CancellationToken? cancellationToken)
 
         var ipAddress = string.IsNullOrEmpty(_config.IPAddress) ? IPAddress.Any : IPAddress.Parse(_config.IPAddress);
         _explicitEndPoint = new ExplicitProxyEndPoint(ipAddress, _config.Port, true);
-        if (!RunTime.IsWindows)
+        // Fired when a CONNECT request is received
+        _explicitEndPoint.BeforeTunnelConnectRequest += OnBeforeTunnelConnectRequest;
+        _explicitEndPoint.GenericCertificate = _proxyServer.CertificateManager.LoadRootCertificate();
+        if (_config.InstallCert)
```
waldekmastykarz, line 105: Shouldn't we keep `!RunTime.IsWindows` here to avoid registering the cert twice? Since it's already being done by Windows as part of `SetAsSystemProxy` on line 124, we only need this for non-Windows OSes

dev-proxy/ProxyEngine.cs
```diff
@@ -119,12 +118,12 @@ public async Task Run(CancellationToken? cancellationToken)
             _logger.LogInfo($"Listening on {endPoint.IpAddress}:{endPoint.Port}...");
         }
 
-        if (RunTime.IsWindows)
+        if (RunTime.IsWindows && _config.AsSystemProxy)
```
waldekmastykarz, line 121: Rather than repeating `&& _config.AsSystemProx` on each if statement, let's wrap the whole `if` block in it, because it applies to all cases, including the generic warning at the end, which should only be displayed if you intend to use Dev Proxy as the system proxy.

waldekmastykarz (changes requested):
Nice! We're getting closer. A couple more comments.

dev-proxy/ProxyEngine.cs
ohaucke: But that would mean that we have either the same condition multiple times, the same else-block twice or missing out on a platform.

```cs
if (_config.AsSystemProxy)
{
    if (RunTime.IsWindows)
    {
        _proxyServer.SetAsSystemHttpsProxy(_explicitEndPoint);
    }
    else if (RunTime.IsMac)
    {
        ToggleSystemProxy(ToggleSystemProxyAction.On, _config.IPAddress, _config.Port);
    }
    else
    {
        _logger.LogWarn("Configure your operating system to use this proxy's port and address");
    }
}
else
{
    _logger.LogWarn("Configure your operating system to use this proxy's port and address");
}
```

```cs
if (RunTime.IsWindows)
{
    if (_config.AsSystemProxy)
    {
        _proxyServer.SetAsSystemHttpsProxy(_explicitEndPoint);
    }
}
else if (RunTime.IsMac)
{
    if (_config.AsSystemProxy)
    {
        ToggleSystemProxy(ToggleSystemProxyAction.On, _config.IPAddress, _config.Port);
    }
}
else
{
    _logger.LogWarn("Configure your operating system to use this proxy's port and address");
}
```

dev-proxy/ProxyEngine.cs
ohaucke: It would only be added once but sure, doesn't hurt.

dev-proxy/ProxyEngine.cs
```diff
@@ -154,6 +153,7 @@ public async Task Run(CancellationToken? cancellationToken)
 
     private void FirstRunSetup()
     {
+        if (!_config.InstallCert) { return; }
```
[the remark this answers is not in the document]
ohaucke: Oh sorry, i forgot about that line - i added it at the beginning so i won't forget that function and then i still forgot about it :D

dev-proxy/ProxyEngine.cs
ohaucke: Or is it something like this, that you have in mind?
```cs

if (_config.AsSystemProxy)
{
    if (RunTime.IsWindows)
    {
        _proxyServer.SetAsSystemHttpsProxy(_explicitEndPoint);
    }
    else if (RunTime.IsMac)
    {
        ToggleSystemProxy(ToggleSystemProxyAction.On, _config.IPAddress, _config.Port);
    }
    else
    {
        _logger.LogWarn("Configure your operating system to use this proxy's port and address");
    }
}
else
{
    _logger.LogInfo("Configure your application to use this proxy's port and address");
}
```

waldekmastykarz: Correct, this latest suggestion is what I had in mind. I like how you added a different info message for configuring the application to use proxy. 👏

waldekmastykarz (changes requested):
It seems like we changed something that we shouldn't have...

After building the code from this PR, when I run proxy with the default options, I'm getting a cert issue, even though it's already trusted on my machine: 

image

If I build the code from main, all is working as intended.

Let's see what's changed and why with these changes, previously trusted certificates no longer work.

ohaucke:
That's odd maybe it's loading the wrong cert for the endpoint
https://github.com/microsoft/dev-proxy/pull/530/files#diff-ce6a87a0599e7d66093279273997c48e9c39c5721aa858b9a99b8ab487311668R104

Unfortunately i only have a linux and windows environment to test.

dev-proxy/ProxyEngine.cs
ohaucke: Now i know, why i did that: `devproxy --as-system-proxy false`
If you run that on windows, it will not install the cert even though it should be installed because the default for `--install-cert` is `true`.

waldekmastykarz:
> That's odd maybe it's loading the wrong cert for the endpoint https://github.com/microsoft/dev-proxy/pull/530/files#diff-ce6a87a0599e7d66093279273997c48e9c39c5721aa858b9a99b8ab487311668R104
> 
> Unfortunately i only have a linux and windows environment to test.

No worries, x-plat dev is hard with the different permutations and specifics. I'll dig into it and report back.

waldekmastykarz:
> That's odd maybe it's loading the wrong cert for the endpoint https://github.com/microsoft/dev-proxy/pull/530/files#diff-ce6a87a0599e7d66093279273997c48e9c39c5721aa858b9a99b8ab487311668R104
> 
> Unfortunately i only have a linux and windows environment to test.

This line seems indeed to be the culprit. When I remove it, all works as intended. I'll dig deeper as to what's special about it

waldekmastykarz:
If I look at the code before this PR, then `_explicitEndPoint.GenericCertificate` stays null after we call `_proxyServer.CertificateManager.EnsureRootCertificate();`. Where did you find out that we need to call `_explicitEndPoint.GenericCertificate = _proxyServer.CertificateManager.LoadRootCertificate();` to load the cert?

waldekmastykarz:
If I remove this line `_explicitEndPoint.GenericCertificate = _proxyServer.CertificateManager.LoadRootCertificate();` altogether, proxy seems to be working just fine with the `--install-cert false --as-system-proxy false` options. Is it needed on Windows?

ohaucke:
I added that line because otherwise `_proxyServer.Start()` would call `CertificateManager.EnsureRootCertificate()`.

```cs
public void Start(bool changeSystemProxySettings = true)
{
    // ...
    SetThreadPoolMinThread(ThreadPoolWorkerThread);
    if (ProxyEndPoints.OfType<ExplicitProxyEndPoint>().Any((ExplicitProxyEndPoint x) => x.GenericCertificate == null))
    {
        CertificateManager.EnsureRootCertificate();
    }
    // ...
}
```

Maybe we could change it to the following?
```cs
if (_config.InstallCert)
{
    _proxyServer.CertificateManager.EnsureRootCertificate();
}
else
{
    _explicitEndPoint.GenericCertificate = _proxyServer.CertificateManager.LoadRootCertificate();
}
```

waldekmastykarz:
> Maybe we could change it to the following?
> 
> ```cs
> if (_config.InstallCert)
> {
>     _proxyServer.CertificateManager.EnsureRootCertificate();
> }
> else
> {
>     _explicitEndPoint.GenericCertificate = _proxyServer.CertificateManager.LoadRootCertificate();
> }
> ```

Let's try. Could you please update your code and see if it works properly on Windows? I'll then verify it on mac. I appreciate your help 👏

ohaucke:
Ok, i've tested it with the changes under windows:
| Parameters | Result |
|---|---|
| `--as-system-proxy true --install-cert true` | no ssl error and set as system proxy |
| `<no args>` | no ssl error and set as system proxy |
| `--as-system-proxy false --install-cert false` | as expected "throws ssl error" and you need to ignore the error |
| `--as-system-proxy false --install-cert true` | doesn't throw ssl error because root cert is installed |
| `--as-system-proxy true --install-cert false` | get's blocked because `--as-system-proxy` requires `--install-cert` |

I've restarted powershell and removed the installed cert after each test.

waldekmastykarz:
Thanks for the detailed tests! I'll run the same on mac to compare the results and will report back

dev-proxy/ProxyEngine.cs
```diff
@@ -95,18 +99,16 @@ public async Task Run(CancellationToken? cancellationToken)
...
+        // Fired when a CONNECT request is received
+        _explicitEndPoint.BeforeTunnelConnectRequest += OnBeforeTunnelConnectRequest;
+        if (_config.InstallCert)
         {
-            // we need to change this to a value lower than 397
-            // to avoid the ERR_CERT_VALIDITY_TOO_LONG error in Edge
-            _proxyServer.CertificateManager.CertificateValidDays = 365;
-            // we need to call it explicitly for non-Windows OSes because it's
-            // a part of the SetAsSystemHttpProxy that works only on Windows
             _proxyServer.CertificateManager.EnsureRootCertificate();
```
waldekmastykarz, line 106: We don't need to call this explicitly on Windows, because it's already taken care of as part of `_proxyServer.Start()`. Let's update the logic to reflect this

waldekmastykarz (changes requested):
I can confirm that it's working the same way on macOS. Let's do one more update before the final check to merge it 👍

dev-proxy/ProxyEngine.cs
ohaucke: If we only rely on `SetAsSystemHttpProxy`, it would break `--as-system-proxy false --install-cert true`.

waldekmastykarz: Ah, you're right. Carry on

waldekmastykarz (approved):

PR #590, opened 2024-02-26
Added configuring for HTTP proxy

excelsiorvita:
Brought back the setting of HTTP proxy on Windows, that was removed in ticket #444.
Extended the toggle-proxy.sh script so that it also enables/disabled HTTP proxy on macOS


excelsiorvita:
@microsoft-github-policy-service agree

waldekmastykarz (approved):
Works like a charm. Nothing to add 👏

PR #592, opened 2024-02-28
Show hotkeys log on proxy start and after clear screen

### LINKED ISSUE #585: Show hotkeys on start and after clearing output
To make hotkeys more discoverable, we should print them on proxy start and after clearing screen:

```text
hotkeys: (r)ecord, (s)top recording, (c)lear screen
```

oleguchok:
It should close #585  

I extracted common method for hotkeys instructions logging
Also I use `_logger` instance to log instructions in `ReadKeys` method instead of `Console.WriteLine` (as far as I understand `_logger` is a wrapper of `Console`)

### DIFF
--- a/dev-proxy/ProxyEngine.cs
+++ b/dev-proxy/ProxyEngine.cs
@@ -580,7 +580,7 @@ public class ProxyEngine
     
     private void LogHotkeysInstructions()
     {
-        _logger.LogInformation("Hotkeys: (r)ecord, (s)top recording, (c)lear screen");
-        _logger.LogInformation("Press CTRL+C to stop Dev Proxy\r\n");
+        Console.WriteLine("Hotkeys: (r)ecord, (s)top recording, (c)lear screen");
+        Console.WriteLine("Press CTRL+C to stop Dev Proxy");
     }
 }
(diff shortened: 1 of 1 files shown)


dev-proxy/ProxyEngine.cs
```diff
@@ -578,4 +577,10 @@ Task OnCertificateSelection(object sender, CertificateSelectionEventArgs e)
         // set e.clientCertificate to override
         return Task.CompletedTask;
     }
+    
+    private void LogHotkeysInstructions()
+    {
+        _logger.LogInformation("Hotkeys: (r)ecord, (s)top recording, (c)lear screen");
```
waldekmastykarz, line 583: If in the future we add the ability to configure different loggers, then we still want these messages to be shown in the terminal. Let's use `Console.WriteLine` rather than the logger.

waldekmastykarz (changes requested):

dev-proxy/ProxyEngine.cs
oleguchok: Got it. Updated

waldekmastykarz (dismissed):

Issue #558, opened 2024-02-12
[BUG]: <Fix for Dev Proxy Installation Script on Linux: Addressing chmod Error and xattr Compatibility Issue>


SilentSobs:
### Description

The Dev Proxy installation script encounters two issues when executed on a Linux system:

Chmod Issue:

Issue: The script attempts to configure devproxy and its files as executable using the chmod command but encounters an error: chmod: cannot access './libe_sqlite3.dylib': No such file or directory.
Expected Behavior: The script should use the correct file name (libe_sqlite3.so) in the chmod command.
Proposed Fix: Update the chmod command to chmod +x ./devproxy ./libe_sqlite3.so.

xattr Issue:

Issue: The script tries to remove the quarantine attribute using the xattr command, which is specific to macOS and not available on Linux.
Expected Behavior: The script should gracefully handle the absence of the xattr command on Linux and print a warning message before skipping the quarantine attribute removal.
Proposed Fix: Check for the existence of the xattr command before attempting to use it. If not found, print a warning and skip the quarantine attribute removal.

### Expected behaviour

After running the Dev Proxy installation script on Linux using the provided command:
```
bash -c "$(curl -sL https://aka.ms/devproxy/setup.sh)"

```
I expect the script to correctly configure devproxy and its files as executable without encountering the chmod: cannot access './libe_sqlite3.dylib': No such file or directory error.
---------------------------------------------------------------------------------------------
Regarding xattr Issue:
Expected Behavior:
When running the script on a Linux system, I expect the script to intelligently handle the absence of the xattr command. Specifically, the script should print a warning message stating that the xattr command is not found and gracefully skip the quarantine attribute removal step.

### Actual behaviour

Regarding chmod Issue:
Actual Behavior:
Upon running the Dev Proxy installation script on Linux with the provided command:
```
bash -c "$(curl -sL https://aka.ms/devproxy/setup.sh)"

```
The script encounters an error during the configuration of devproxy and its files as executable. The specific error message is:
```
chmod: cannot access './libe_sqlite3.dylib': No such file or directory

```

Regarding xattr Issue:
Actual Behavior:
When running the script on a Linux system, the script attempts to use the xattr command to remove the quarantine attribute. However, the xattr command is specific to macOS and is not available on Linux. As a result, the script fails during this step.

### Steps to reproduce

```
bash -c "$(curl -sL https://aka.ms/devproxy/setup.sh)"

```

### Dev Proxy Version

-

### Operating system (environment)

[...]

PR #608, opened 2024-03-13
Fix for Dev Proxy Installation Script on Linux:  #558

SilentSobs:
Addressing issue https://github.com/microsoft/dev-proxy/issues/558 #558 

To correct and improve the line for version detection that wasn't working properly, and to incorporate the use of grep 
```
version=$(curl -s https://api.github.com/repos/microsoft/dev-proxy/releases/latest | grep -o '"tag_name": *"[^"]*"' | sed 's/"tag_name": *"\(.*\)"/\1/')
```
This version uses grep to filter the line that contains "tag_name":, making it easier to find the version string. 

```
# Check OS and set permissions accordingly
if [ "$(uname)" == "Darwin" ]; then
    # Set permissions for macOS
    chmod +x ./devproxy ./libe_sqlite3.dylib
elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
    # Set permissions for Linux
    chmod +x ./devproxy ./libe_sqlite3.so
fi
```


SilentSobs:
The reason I changed this because while fixing the issue  i was having an issue with fetching version from the previous code 
```
version=$(curl -s https://api.github.com/repos/microsoft/dev-proxy/releases/latest | awk -F: '/"tag_name"/ {print $2}' | sed 's/[", ]//g')
```
But now i can check it is working But at that time it was not able to download the dev proxy from the script.

waldekmastykarz:
> But now i can check it is working But at that time it was not able to download the dev proxy from the script.

Please do. I'll wait with merging until I hear back from you

SilentSobs:
It's working.
image
image

waldekmastykarz:
Is it the existing or the updated version of the download logic working?

SilentSobs:
Updated version.

waldekmastykarz (changes requested):
Works like a charm! Could you also please update the setup-beta.sh script?

SilentSobs:
On macOS (Darwin), the sed -i command expects an argument immediately following -i. If we want to perform in-place editing without creating backup files, we need to provide an empty string as the argument, like this: sed -i ''.

However, on Linux systems, sed -i requires an argument specifying the suffix for backup files. If we want to perform in-place editing without creating backup files, we should omit the suffix altogether, like this: sed -i.

SilentSobs:
image

image

SilentSobs:
Once this installation script is fixed I will start working on the upgrade part which is left . I don't have mac os so it creates an issue for me to test on that environment . Is there any workaround for that so I can test the application on mac os something like a virtual environment for a developer.

waldekmastykarz (approved):

waldekmastykarz:
> Once this installation script is fixed I will start working on the upgrade part which is left . I don't have mac os so it creates an issue for me to test on that environment . Is there any workaround for that so I can test the application on mac os something like a virtual environment for a developer.

You could try using a GitHub macOS agent and then use GitHub Actions to test your logic.

waldekmastykarz:
Awesome work handling the discrepancies between OSes! Thank you!

PR #620, opened 2024-03-22
Adds abstractions to release artifacts. Closes #613

### LINKED ISSUE #613: Add `dev-proxy-abstractions.dll` to release artifacts
We should publish the `dev-proxy-abstractions.dll` along our release artifacts to make it easier for folks to build custom plugins. Right now, the only way for them to get the assembly is to clone the repo and build the project themselves which is unnecessarily complicated.

waldekmastykarz:
Adds abstractions to release artifacts. Closes #613


garrytrinder (dismissed):

waldekmastykarz:
There's possibly a simpler way, one that doesn't require publishing a separate assembly for each OS. Instead, we'd have one that can be used on any OS. I'll check it out and update the PR if needed

waldekmastykarz:
Updated the PR with a simplified workflow. @garrytrinder would you mind having another look?

PR #625, opened 2024-03-28
Fix for detecting OS and architecture

rutgersmit:
Updated the detecting of the OS and architecture using $env:OS and [Environment]::Is64BitOperatingSystem.

Tested on virtualised Windows 11, host is MacOS (Silicon M1).

### DIFF
--- a/scripts/setup.ps1
+++ b/scripts/setup.ps1
@@ -40,10 +40,11 @@ $base_url = "https://github.com/microsoft/dev-proxy/releases/download/$version/d
 
 # Check system architecture
 $os = $env:OS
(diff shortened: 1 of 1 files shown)


waldekmastykarz (changes requested):
Hey @rutgersmit, thanks for the submission. Unfortunately, it doesn't work. On macOS in PowerShell 7.4.1, `$env:OS` doesn't return anything, so we can't use it as a replacement for the current code.

waldekmastykarz:
We'll use #626 which seems to fix the issue in a way that's compatible with Windows PowerShell and PowerShell Core. We appreciate your help

PR #626, opened 2024-03-28
Updates setup files. Fixes #624

### LINKED ISSUE #624: PowerShell setup script fails in Windows PowerShell
When you try to run the PowerShell setup script in Windows PowerShell it fails with `Unsupported OS . Aborting`. This is caused by Windows PowerShell not supporting `$PSVersionTable.OS` which we use to detect the OS. We need to either update the setup script to use an alternative way of detecting the OS or show a clear error message that Windows PowerShell is not supported and you should use Microsoft PowerShell instead. If we can't support Windows PowerShell, we should also include it in our docs.

waldekmastykarz:
Updates setup files. Fixes #624


garrytrinder (approved):
✅

image

PR #644, opened 2024-04-12
Updates version to v0.16.2

garrytrinder:


waldekmastykarz (changes requested):
Let's update the .iss file as well

waldekmastykarz (approved):

PR #652, opened 2024-04-17
Adds Windows beta installer

garrytrinder:


.github/workflows/create-release.yml
```diff
@@ -66,16 +66,23 @@ jobs:
         with:
           name: binaries-dev-proxy-abstractions-${{ github.ref_name }}
           path: ./dev-proxy-abstractions-${{ github.ref_name }}.zip
-      - name: Add install file
-        if: matrix.architecture == 'win-x64' || matrix.architecture == 'win-x86'
-        run: cp ./install.iss ./${{ env.release }}
       - name: Add installer icon
         if: matrix.architecture == 'win-x64' || matrix.architecture == 'win-x86'
         run: cp ./media/icon.ico ./${{ env.release }}
+      - name: Add install file
```
waldekmastykarz, line 72: Let's simplify the logic between a regular and a beta install with a variable to which we assign the installer file name. Then, if we're running a beta release, we'll reassign it to the beta installer. That way, we won't have to repeat the installation commands and maintain them in multiple places.

waldekmastykarz (changes requested):

.github/workflows/create-release.yml
```diff
@@ -66,16 +66,23 @@ jobs:
...
+        if: (matrix.architecture == 'win-x64' || matrix.architecture == 'win-x86') && !contains(github.ref_name, 'beta')
```
waldekmastykarz, line 73: Now that I've seen the `contains` check you've added, let's simplify `matrix.architecture == 'win-x64' || matrix.architecture == 'win-x86'` to `contains(matrix.architecture, 'win-')`

waldekmastykarz (approved):

PR #659, opened 2024-04-18
Adds outdated command. Closes #657

### LINKED ISSUE #657: `devproxy outdated` command
So we don't replicate the same logic in the extension to check if there is a new version of proxy, we should expose the logic already implemented in Dev Proxy behind a command that can be executed.

The command should return either an empty string (no updated needed), or the new version that the user should upgrade to.

garrytrinder:
Closes #657

### DIFF
--- a/dev-proxy/OutdatedCommandHandler.cs
+++ b/dev-proxy/OutdatedCommandHandler.cs
@@ -1,0 +1,13 @@
+﻿using Microsoft.Extensions.Logging;
+
+namespace Microsoft.DevProxy;
+
+public static class OutdatedCommandHandler
+{
+    public static async Task CheckVersion(ILogger logger)
+    {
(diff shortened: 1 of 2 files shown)


dev-proxy/OutdatedCommandHandler.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Microsoft.DevProxy;
+
+public static class OutdatedCommandHandler
+{
+    public static async Task CheckVersion()
+    {
+        var releaseInfo = await UpdateNotification.CheckForNewVersion(ProxyCommandHandler.Configuration.NewVersionNotification);
+        Console.WriteLine(releaseInfo?.Version);
```
waldekmastykarz, line 8: Let's use logger instead of writing directly to Console in case we introduce more/other loggers in the future

waldekmastykarz (changes requested):

waldekmastykarz (approved):

PR #688, opened 2024-05-04
Added Timestamp to Log

SilentSobs:


waldekmastykarz (changes requested):
Nice! Before we continue, let's clean up instances where we've added logging time manually, like here: https://github.com/microsoft/dev-proxy/blob/0a11bd698b76fc44b07f2fa88c2c26bf1865c7be/dev-proxy-plugins/RequestLogs/ApiCenterProductionVersionPlugin.cs#L118 so that we're not showing it twice.

waldekmastykarz:
Hey @SilentSobs, is the PR ready for review? If so, let's mark is as such

SilentSobs:
> Hey @SilentSobs, is the PR ready for review? If so, let's mark is as such

Yes.

waldekmastykarz (approved):

PR #714, opened 2024-05-15
Reporters

waldekmastykarz:
This PR shows a concept of a new architecture for request log plugins to share their reports. The idea is, that each plugin produces an object that represents its report and adds it to a central location. Then, one or more transformers can change the shape of the report. Finally, one or more reporters output the data. Using this feature, we get more flexibility for different use cases of Dev Proxy, like interactive in console, or integrated in CI where the output is automatically processed.

Transformers and reporters are implemented as request log plugins, but instead of processing request logs they work on the prepared reports.

Some considerations:
- right now, if you don't define a transformer and reporter, you won't see the result of a request log plugin. Do we want to keep this, or should each plugin output for example by default to console and optionally to reporters?
- right now, we separated transformers and reporters. That way, we can combine them, eg. JSON output to console, JSON output to file, etc. Is this correct, or would we rather have them combined and done in such a way, that reporters only read and output data but don't modify it (like a transformer), so that it's possible to use multiple reporters?


waldekmastykarz:
> Some considerations:
> 
> * right now, if you don't define a transformer and reporter, you won't see the result of a request log plugin. Do we want to keep this, or should each plugin output for example by default to console and optionally to reporters?
> * right now, we separated transformers and reporters. That way, we can combine them, eg. JSON output to console, JSON output to file, etc. Is this correct, or would we rather have them combined and done in such a way, that reporters only read and output data but don't modify it (like a transformer), so that it's possible to use multiple reporters?

Thinking about it some more, I think it would be better to:

- have each plugin display something by default. Perhaps not the full report but at least some information that confirms that it's done what it's supposed to do
- combine transformers and reporters
- leave original report objects unaltered so that they can be processed by multiple reporters
- drop the idea of a logger reporter, and have reporters store their output in files. Given that stdout is filled with other Dev Proxy information, it's impractical to extract the structured reports out of it. Having each report in a separate file is more convenient and robust
- write output to a file following the report_reporter.ext pattern. If the file already exists, we append current date and time to it, like report_reporter_yyyyMMdd_hhmmss.ext. If you run Dev Proxy in CI, you get a predictable file name that you can easily get to for further processing. If you run Dev Proxy interactively and start multiple recordings, you avoid reports being overwritten

waldekmastykarz:
I submitted a second PR that shows the latest structure as proposed in https://github.com/microsoft/dev-proxy/pull/714#issuecomment-2114161502

PR #715, opened 2024-05-15
Refactors IProxyLogger to ILogger

waldekmastykarz:
This PR proposes refactoring of IProxyLogger to the standard .NET ILogger. It uses a custom ConsoleLogFormatter to keep our custom formatting and adds an extra RequestLogger to keep raising events when a request is logged. The idea behind this change is to minimize the amount of custom implementation and use as standard .net features when possible.


waldekmastykarz:
Closing in favor of #716

PR #741, opened 2024-05-23
Update install scripts to include ChangesEnvironment flag

garrytrinder:


scripts/local-setup.ps1
```diff
@@ -1,4 +1,4 @@
-$versionString = "v0.18.0-beta.1"
+$versionString = "v0.18.0-beta.env"
```
waldekmastykarz, line 1: Is this change intentional?

waldekmastykarz (changes requested):

scripts/local-setup.ps1
garrytrinder: No, good catch.

waldekmastykarz (approved):

PR #760, opened 2024-06-04
#694 log errors when DevTools cannot be started

### LINKED ISSUE #694: Warn when browser is already running before starting DevTools
### Description

I cannot make devtools to work, even though there are no errors logged in devproxy

### Expected behaviour

opening `http://localhost:9222/devtools/inspector.html?ws=localhost:62690` opens devtools

### Actual behaviour

Connection refused. Nothing listening on 9222?

### Steps to reproduce

`devproxy -c "C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\presets\microsoft-graph.json" --record --log-level Trace`

with
```json
  "plugins": [
      {
        "name": "DevToolsPlugin",
        "enabled": true,
        "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
        "configSection": "devTools"
      },
```

### Dev Proxy Version

0.17.1

### Operating system (environment)

Windows

### Shell

PowerShell

### Configuration file

```json
{
  "$schema": "https://raw.githubusercontent.com/microsoft/dev-proxy/main/schemas/v0.17.1/rc.schema.json",
  "plugins": [
      {
        "name": "DevToolsPlugin",
        "enabled": true,
        "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
        "configSection": "devTools"
      },
    {
      "name": "GraphSelectGuidancePlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll"
    },
    {
      "name": "GraphBetaSupportGuidancePlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
      "urlsToWatch": [
        "https://graph.microsoft.com/beta/*",
        "https://graph.microsoft.us/beta/*",
        "https://dod-graph.microsoft.us/beta/*",
        "https://microsoftgraph.chinacloudapi.cn/beta/*"
      ]
    },
    {
      "name": "GraphConnectorGuidancePlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
      "urlsToWatch": [
        "https://graph.microsoft.com/*/external/connections/*/schema",
        "https://graph.microsoft.us/*/external/connections/*/schema",
        "https://dod-graph.microsoft.us/*/external/connections/*/schema",
        "https://microsoftgraph.chinacloudapi.cn/*/external/connections/*/schema"
      ]
    },
    {
      "name": "GraphSdkGuidancePlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll"
    },
    {
      "name": "ODataPagingGuidancePlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll"
    },
    {
      "name": "GraphClientRequestIdGuidancePlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll"
    },
    {
      "name": "GraphRandomErrorPlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
      "configSection": "graphRandomErrorsPlugin"
    },
    {
      "name": "ExecutionSummaryPlugin",
      "enabled": true,
      "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
      "configSection": "executionSummaryPlugin"
    }
  ],
  "urlsToWatch": [
    "https://graph.microsoft.com/v1.0/*",
    "https://graph.microsoft.com/beta/*",
    "https://graph.microsoft.us/v1.0/*",
    "https://graph.microsoft.us/beta/*",
    "https://dod-graph.microsoft.us/v1.0/*",
    "https://dod-graph.microsoft.us/beta/*",
    "https://microsoftgraph.chinacloudapi.cn/v1.0/*",
    "https://microsoftgraph.chinacloudapi.cn/beta/*"
  ],
  "graphRandomErrorsPlugin": {
    "allowedErrors": [ 429, 500, 502, 503, 504, 507 ]
  },
  "executionSummaryPlugin": {
    "groupBy": "url"
  },
  "devTools": {
    "preferredBrowser": "Chrome"
  },
  "rate": 50,
  "labelMode": "text",
  "logLevel": "information"
}
```

### Additional Info

Log output:
```
PS C:\Users\piotrkarpala> devproxy -c "C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\presets\microsoft-graph.json" --record --log-level Trace
Loading plugin DevToolsPlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin DevToolsPlugin...
DevTools available at http://localhost:9222/devtools/inspector.html?ws=localhost:62690
Plugin DevToolsPlugin registered.
Loading plugin GraphSelectGuidancePlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin GraphSelectGuidancePlugin...
Plugin GraphSelectGuidancePlugin registered.
Loading plugin GraphBetaSupportGuidancePlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin GraphBetaSupportGuidancePlugin...
Plugin GraphBetaSupportGuidancePlugin registered.
Loading plugin GraphConnectorGuidancePlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin GraphConnectorGuidancePlugin...
Plugin GraphConnectorGuidancePlugin registered.
Loading plugin GraphSdkGuidancePlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin GraphSdkGuidancePlugin...
Plugin GraphSdkGuidancePlugin registered.
Loading plugin ODataPagingGuidancePlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin ODataPagingGuidancePlugin...
Plugin ODataPagingGuidancePlugin registered.
Loading plugin GraphClientRequestIdGuidancePlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin GraphClientRequestIdGuidancePlugin...
Plugin GraphClientRequestIdGuidancePlugin registered.
Loading plugin GraphRandomErrorPlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin GraphRandomErrorPlugin...
Plugin GraphRandomErrorPlugin registered.
Loading plugin ExecutionSummaryPlugin from: C:\Users\piotrkarpala\AppData\Local\Programs\Dev Proxy\plugins\dev-proxy-plugins.dll
Registering plugin ExecutionSummaryPlugin...
Plugin ExecutionSummaryPlugin registered.
Listening on 127.0.0.1:8000...
Set endpoint at Ip 127.0.0.1 and port: 8000 as System HTTP Proxy
Set endpoint at Ip 127.0.0.1 and port: 8000 as System HTTPS Proxy
Hotkeys: issue (w)eb request, (r)ecord, (s)top recording, (c)lear screen
Press CTRL+C to stop Dev Proxy

◉ Recording...
 request     GET https://graph.microsoft.com/v1.0/users
 warning   ╭ To help Microsoft investigate errors, to each request to Microsoft Graph
           │ add the client-request-id header with a unique GUID.
           │ More info at https://aka.ms/devproxy/guidance/client-request-id
           ╰ GET https://graph.microsoft.com/v1.0/users
```

karpikpl:
This fixes #694 

I'm not sure if we should add support for unit testing and Dispose to cleanup after DevTools?


dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -23,6 +23,12 @@ public enum PreferredBrowser
 public class DevToolsPluginConfiguration
 {
     public PreferredBrowser PreferredBrowser { get; set; } = PreferredBrowser.Edge;
+
+    /// <summary>
+    /// Path to the browser executable.
+    /// If not set, the plugin will try to find the browser executable based on the PreferredBrowser.
+    /// </summary>
+    public string PreferredBrowserPath { get; set; } = string.Empty;
```
waldekmastykarz, line 31: Why do we need to ask folks for the path to the preferred browser? It seems excessive to ask for both the preferred browser and the path, as either one should be sufficient. Is there a specific scenario for which we need this?

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -83,6 +102,25 @@ private void InitInspector()
...
+            Logger.LogError("Browser executable not found at {processName}", processName);
+            return;
+        }
+
+        // find if the process is already running
+        var processes = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(processName));
+
+        if (processes.Length > 0)
+        {
+            Logger.LogError(@"Found existing process {processName} with Id {processId}.
```
waldekmastykarz, line 118: Let's write this message in one line

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -83,6 +102,25 @@ private void InitInspector()
...
+        }
+
+        // find if the process is already running
+        var processes = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(processName));
+
+        if (processes.Length > 0)
+        {
+            Logger.LogError(@"Found existing process {processName} with Id {processId}. 
+            Could not start DevTools plugin. 
+            Please close existing browser windows before starting DevTools plugin.", processName, processes[0].Id);
```
waldekmastykarz, line 120: Let's print all PIDs in a comma-separated list, not just the first one.

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -83,6 +102,25 @@ private void InitInspector()
...
+            return;
+        }
+
+        // find if the process is already running
+        var processes = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(processName));
+
+        if (processes.Length > 0)
+        {
+            Logger.LogError(@"Found existing process {processName} with Id {processId}. 
+            Could not start DevTools plugin.
```
waldekmastykarz, line 119: Let's replace the `DevTools plugin` string with the value of the `Name` property for consistency.

waldekmastykarz (changes requested):
Great start! Let's check a couple of things before we continue

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
karpikpl: so the feature doesn't work on Linux since there's no browser path there - or I cannot find the browser.
If I install chrome and provide the path to it, it works.

So we should either throw an error if someone tries to run it in Ubuntu (like I'm using it) or find a way to find chrome there

waldekmastykarz: Understood. Is there a standard location where the browser install to on Linux? If so, we could update the logic in here to keep it consistent with other OSes: https://github.com/microsoft/dev-proxy/blob/9e73c68c7d85780d0820d745d0dd020cfff0f0c3/dev-proxy-plugins/Inspection/DevToolsPlugin.cs#L62-L76

karpikpl: I don't think it is. I've tested with binary installed using playwright. 
Someone would have to add chrome to PATH, then we could reference it using just executable name.

This isn't the case for me on ubuntu on WSL, which kind of make sense, since it's just terminal

waldekmastykarz: I don't think that Playwright and WSL are a good comparison here. They're both headless. When you're working in WSL, you wouldn't run the browser in it, but rather start it on the host, right? Shouldn't we be checking for regular browser processes rather than processes temporarily spawned by Playwright or other similar tools?

karpikpl: so yes and no.
I do most work on WSL - so that's when I need to setup proxy. It's a pain to go WSL -> windows host but not other way around. 
Even though WSL is headless - you can still open chrome, run playwright etc.

That `PreferredBrowserPath` is not a hill I want to die on :) 
I don't like the idea of providing path to browser as much as anyone else, but it was the only thing that worked for me on WSL. 
On the Ubuntu VM chrome might be in a system path, so I can just make that assumption and for everything non-OSX/Windows check for chrome binary in path and fail if it's not there? Then I could remove `PreferredBrowserPath`  and on WSL update my bash setting to include it?

here how it looks right now on WSL with preferred path:
image

waldekmastykarz: I'm not against `PreferredBrowserPath` on principle, just trying to find a way to get it all to work with as little friction as possible 🙂  If we could provide a default path with the ability to override then it would be a nice middle ground between a default setup that works for most and an escape hatch in case you need something else.

Is there a reason you install the browser in WSL using Playwright? I'd expect the browser to be installed as an OS-wide app using a package manager like apt, but perhaps there are some good reasons to do it like you do.

I appreciate your feedback!

karpikpl: I've updated the code with default browser path and added comment for install instructions

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -61,32 +68,67 @@ private static int GetFreePort()
...
-                ? Environment.ExpandEnvironmentVariables(@"%ProgramFiles(x86)%\Microsoft\Edge\Application\msedge.exe")
-                : "/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge"
+            PreferredBrowser.Chrome => RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? Environment.ExpandEnvironmentVariables(@"%ProgramFiles%\Google\Chrome\Application\chrome.exe")
+                : RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+                : RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "/usr/bin/google-chrome"
+                : throw new NotSupportedException("Unsupported operating system."),
+            PreferredBrowser.EdgeDev => RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? Environment.ExpandEnvironmentVariables(@"%ProgramFiles(x86)%\Microsoft\Edge Dev\Application\msedge.exe")
+                : RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "/Applications/Microsoft Edge Dev.app/Contents/MacOS/Microsoft Edge Dev"
+                : RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "/usr/bin/microsoft-edge-dev"
```
waldekmastykarz, line 87: Seems like we lost Edge

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -61,32 +68,67 @@ private static int GetFreePort()
...
+        string processName = string.Empty;
+
+        try
+        {
+            processName = GetChromiumProcessName(_configuration);
+        }
+        catch (NotSupportedException ex)
+        {
+            Logger.LogError(ex, "Error starting {name}. Error finding the browser.", Name);
+        }
```
waldekmastykarz, line 104: We should stop early here

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -61,32 +68,67 @@ private static int GetFreePort()
...
+
+        // check if the browser is installed
+        if (!File.Exists(processName))
+        {
+            Logger.LogError("Error starting {name}. Browser executable not found at {processName}", Name, processName);
+            return;
+        }
+
+        // find if the process is already running
+        var processes = Process.GetProcessesByName(Path.GetFileNameWithoutExtension(processName));
```
waldekmastykarz, line 114: This doesn't work on macOS because the actual process name is stored in `Process.MainModule.ModuleName`, so we need to update this logic to differentiate across OSes

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
karpikpl: that's right,  sorry for that. GH doesn't really show those changes well.

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
karpikpl: ok, I'll dust off my MPro and check.

waldekmastykarz: All good, I've done a couple of updates already

waldekmastykarz:
I've updated the PR and merged it manually. I've done some changes as to how we detect browsers to accommodate different versions of Edge across OSes. Check out the final commit at: https://github.com/microsoft/dev-proxy/commit/2d583e302848ebbc2bfa5b6d5887c85cfa595721.

Thank you so much for working with us on this!

PR #804, opened 2024-06-20
Update command alias. Closes #803

### LINKED ISSUE #803: Update setup scripts to use correct CLI command
We currently use `aad app add` command in CLI for Microsoft 365 to setup Microsoft Entra ID app registration for the sample app. We need to update this to use the new command alias `entra app add`.

image

garrytrinder:
Closes #803

### DIFF
--- a/samples/scripts/setup.ps1
+++ b/samples/scripts/setup.ps1
@@ -3,8 +3,8 @@ Write-Output "Sign in to Microsoft 365..."
 npx -p @pnp/cli-microsoft365 -- m365 login --authType browser
 
 # create AAD app
-Write-Output "Creating AAD app..."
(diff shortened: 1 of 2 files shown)


samples/scripts/setup.ps1
```diff
@@ -3,8 +3,8 @@ Write-Output "Sign in to Microsoft 365..."
 npx -p @pnp/cli-microsoft365 -- m365 login --authType browser
 
 # create AAD app
-Write-Output "Creating AAD app..."
-$appId = npx -p @pnp/cli-microsoft365 -- "m365 aad app add --name graph-developer-proxy-samples --multitenant --redirectUris http://localhost:3000/withsdk.html,http://localhost:3000/nosdk.html --apisDelegated https://graph.microsoft.com/User.Read.All,https://graph.microsoft.com/Presence.Read.All --grantAdminConsent --platform spa --query appId -o text"
+Write-Output "Creating Entra ID app..."
```
waldekmastykarz, line 6: It's actually `Entra app`. You only use `Entra ID` when referring to the product itself, but not its features

waldekmastykarz (changes requested):

samples/scripts/setup.ps1
garrytrinder: Good catch. Naming is hard 😅

waldekmastykarz (approved):

PR #806, opened 2024-06-24
Add new config command to easily open devproxyrc.json

mwiedemeyer:
I would like to add a new command `config` which automatically opens the `devproxyrc.json` in the default editor.
This would make it a lot easier for new developers to find the file.
The [documentation here](https://learn.microsoft.com/en-us/microsoft-cloud/dev/dev-proxy/get-started?tabs=automated&pivots=client-operating-system-windows#update-the-urls-to-watch) does not mention how to find the file. Instead it only says:
> In the Dev Proxy installation folder, open devproxyrc.json in a text editor.

So with the new command it would be a lot easier, because the documentation can say:
> Run `devproxy config` to open devproxyrc.json in a text editor

### DIFF
--- a/dev-proxy/ProxyHost.cs
+++ b/dev-proxy/ProxyHost.cs
@@ -6,7 +6,6 @@ using Microsoft.Extensions.Logging;
 using System.CommandLine;
 using System.Diagnostics;
 using System.Net;
-using System.Reflection;
 
 namespace Microsoft.DevProxy;
 
--- a/dev-proxy/ProxyHost.cs
+++ b/dev-proxy/ProxyHost.cs
@@ -330,7 +330,7 @@ internal class ProxyHost
         var configCommand = new Command("config", "Open devproxyrc.json");
         configCommand.SetHandler(() =>
         {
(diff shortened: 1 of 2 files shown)


waldekmastykarz:
Thank you for this PR. It'll definitely make it easier to update the default config. We've been hesitant to add in the past because it depends on the developer's file association, but thinking about it some more, it's a fair game to assume that folks have file associations configured on their machine. We'll process it asap. Thanks again!

mwiedemeyer:
Awesome, thank you.

dev-proxy/ProxyHost.cs
```diff
@@ -4,7 +4,9 @@
 using Microsoft.DevProxy.Abstractions;
 using Microsoft.Extensions.Logging;
 using System.CommandLine;
+using System.Diagnostics;
 using System.Net;
+using System.Reflection;
```
waldekmastykarz: We don't need this

waldekmastykarz (approved):
Works like a charm!

Issue #840, opened 2024-08-01
[BUG]: DevToolsPlugin returns the error Unable to enumerate the process modules


qmatteoq:
### Description

When you enable the DevToolsPlugin and you run devproxy, the execution fails with an exception

### Expected behaviour

The browser configured in the devproxyrc.json file starts up with the Developer Tools turned on, showing the requests intercepted by Dev Proxy

### Actual behaviour

DevProxy fails to launch with the following exception

```text
Unhandled exception. System.ComponentModel.Win32Exception (0x80004005): Unable to enumerate the process modules.
   at System.Diagnostics.NtProcessManager.GetModules(Int32 processId, Boolean firstModuleOnly)
   at System.Diagnostics.NtProcessManager.GetFirstModule(Int32 processId)
   at Microsoft.DevProxy.Plugins.Inspection.DevToolsPlugin.<>c__DisplayClass10_0.<GetBrowserProcesses>b__0(Process p)
   at System.Linq.Enumerable.WhereArrayIterator`1.ToArray()
   at Microsoft.DevProxy.Plugins.Inspection.DevToolsPlugin.GetBrowserProcesses(String browserPath)
   at Microsoft.DevProxy.Plugins.Inspection.DevToolsPlugin.InitInspector()
   at Microsoft.DevProxy.Plugins.Inspection.DevToolsPlugin.Register()
   at Microsoft.DevProxy.PluginLoader.LoadPlugins(IPluginEvents pluginEvents, IProxyContext proxyContext)
   at Program.<Main>$(String[] args)
   at Program.<Main>(String[] args)
```

### Steps to reproduce

1) Open devproxyrc.json and add the DevToolsPlugin in the plugins section:

```json
{
  "name": "DevToolsPlugin",
  "enabled": true,
  "pluginPath": "~appFolder/plugins/dev-proxy-plugins.dll",
  "configSection": "devTools"
}
```

2) At the end of the file, add a section to define the configuration:

```json
"devTools": {
  "preferredBrowser": "Edge"
}
```

3) Launch DevProxy
4) Observe the exception being triggered

### Dev Proxy Version

0.19.1

### Operating system (environment)

Windows

### Shell

PowerShell

### Configuration file

```json
{
  "$schema": "https://raw.githubusercontent.com/microsoft/dev-proxy/main/schemas/v0.19.1/rc.schema.json",
  "plugins": [
    {
      "name": "ExecutionSummaryPlugin",
      "enabled": true,
[...]

PR #843, opened 2024-08-06
Fixed the process detection logic for Windows for the DevTools plugin

qmatteoq:
This PR addresses issue #840 by improving the logic which detects the browser process in Windows.
In short, Windows has lot of processes (like System or Idle) that can't be accessed by a regular application. You are able to retrieve them, but you aren't allowed to query properties like the Name, otherwise you will get an exception with Access Denied.

The new logic wraps the process iteration process inside a try / catch statement, so that these special processes are ignored.

The PR also adds support for Edge Beta, in addition to Edge and Edge Dev.

### DIFF
--- a/dev-proxy-plugins/Inspection/DevToolsPlugin.cs
+++ b/dev-proxy-plugins/Inspection/DevToolsPlugin.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.DevProxy.Abstractions;
 using Microsoft.DevProxy.Plugins.Inspection.CDP;
 using Microsoft.Extensions.Logging;
+using System.ComponentModel;
 
 namespace Microsoft.DevProxy.Plugins.Inspection;
 
@@ -17,7 +18,8 @@ public enum PreferredBrowser
 {
     Edge,
     Chrome,
-    EdgeDev
+    EdgeDev,
+    EdgeBeta
 }
 
 public class DevToolsPluginConfiguration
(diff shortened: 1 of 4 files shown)


dev-proxy-plugins/Inspection/DevToolsPlugin.cs
```diff
@@ -113,9 +118,37 @@ private string GetBrowserPath(DevToolsPluginConfiguration configuration)
 
     private Process[] GetBrowserProcesses(string browserPath)
     {
-        return Process.GetProcesses().Where(p =>
-            p.MainModule is not null && p.MainModule.FileName == browserPath
-        ).ToArray();
+        List<Process> processes = new List<Process>();
```
waldekmastykarz, line 121: I suggest we simplify it to something like:

```cs
    private Process[] GetBrowserProcesses(string browserPath)
    {
        return Process.GetProcesses().Where(p => {
            try
            {
                return p.MainModule is not null && p.MainModule.FileName == browserPath;
            }
            catch (Exception ex)
            {
                Logger.LogDebug("Error while checking process: {Ex}", ex.Message);
                return false;
            }
        }).ToArray();
    }
```

Since we're not doing anything with the caught exceptions, there's no point in keeping them around. For debugging, we can show the message inline.

waldekmastykarz (changes requested):

dev-proxy-plugins/Inspection/DevToolsPlugin.cs
qmatteoq: Yes, I agree, if you aren't interested in tracking the exception, your solution is cleaner and simpler 😊

waldekmastykarz (approved):

PR #850, opened 2024-08-19
Adds Dev Proxy API

waldekmastykarz:


dev-proxy-abstractions/IProxyConfiguration.cs
```diff
@@ -18,13 +18,18 @@ public enum LabelMode
 
 public interface IProxyConfiguration
 {
-    int Port { get; }
+    int ApiPort { get; }
```
waldekmastykarz, line 21: Sorted properties alphabetically to simplify lookup and management going forward

garrytrinder (approved):

PR #891, opened 2024-09-23
Add jwt create command and createJwtToken endpoint. Closes #882

### LINKED ISSUE #882: Create JWT token command
When simulating auth developers need to generate a fake auth token to use in their code. This is not a trivial task.

We should help developers by creating a command that can generate tokens to be used for testing purposes.

[dotnet user-jwts](https://learn.microsoft.com/aspnet/core/security/authentication/jwt-authn?view=aspnetcore-8.0&tabs=windows) is a command line tool that can generate tokens for testing.

Example: 

```
dotnet user-jwts create --audience 5fd6c579-77d2-4ad4-9064-b8b2297a6fb5 --scope "Posts.Read" --issuer "https://login.microsoftonline.com/bf805346-44e7-43f7-8e60-c5e24aefe49c/v2.0"
```

We should think about integrating this tool in Dev Proxy for convenience.

garrytrinder:


dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,53 @@ public RootCommand GetRootCommand(ILogger logger)
...
+        jwtCreateCommand.AddOption(jwtNameOption);
+        var jwtAudienceOption = new Option<string>("--audience", "The audiences to create the token for.");
+        jwtAudienceOption.AddAlias("-a");
+        jwtAudienceOption.SetDefaultValue("https://myserver.com");
+        jwtCreateCommand.AddOption(jwtAudienceOption);
+        var jwtIssuerOption = new Option<string>("--issuer", "The issuer of the token.");
+        jwtIssuerOption.AddAlias("-i");
+        jwtIssuerOption.SetDefaultValue("dev-proxy");
+        jwtCreateCommand.AddOption(jwtIssuerOption);
+        var jwtRolesOption = new Option<string>("--roles", "A role claim to add to the token. Specify once for each scope. Use a comma to separate scopes.");
```
waldekmastykarz, line 368: Instead of using comma-separated values, let's use the native support for multiple values. That way we don't need parse the string manually and there's less room for error

dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,53 @@ public RootCommand GetRootCommand(ILogger logger)
...
+        jwtRolesOption.AddAlias("-r");
+        jwtCreateCommand.AddOption(jwtRolesOption);
+        var jwtScopesOption = new Option<string>("--scopes", "A scope claim to add to the token. Specify once for each scope. Use a comma to separate roles.");
```
waldekmastykarz, line 371: Instead of using comma-separated values, let's use the native support for multiple values. That way we don't need parse the string manually and there's less room for error

dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,53 @@ public RootCommand GetRootCommand(ILogger logger)
...
+        jwtScopesOption.AddAlias("-s");
+        jwtCreateCommand.AddOption(jwtScopesOption);
+        var jwtClaimsOption = new Option<string>("--claims", "Claims to add to the token. Specify once for each claim in the format \"name=value\". Use a comma to separate claims.");
+        jwtClaimsOption.AddAlias("-c");
+        jwtCreateCommand.AddOption(jwtClaimsOption);
+        var jwtValidForOption = new Option<double>("--valid-for", "The duration for which the token is valid. Duration is set in hours.");
```
waldekmastykarz, line 377: Let's change the unit to minutes. That way we avoid decimals in case someone wants to have a token for 30 or 90 mins

dev-proxy/CommandHandlers/JwtCommandHandler.cs
```diff
@@ -0,0 +1,146 @@
...
+
+    public IEnumerable<string> Roles { get; set; } = [];
+
+    public IDictionary<string, string> CustomClaims { get; set; } = new Dictionary<string, string>();
+
+    public static Jwt Create(
+        string scheme,
+        JwtSecurityToken token,
+        string encodedToken,
+        IEnumerable<string> scopes = null,
```
waldekmastykarz, line 124: Let's address all these build issues before we merge

dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,53 @@ public RootCommand GetRootCommand(ILogger logger)
...
+        jwtIssuerOption.AddAlias("-i");
+        jwtIssuerOption.SetDefaultValue("dev-proxy");
+        jwtCreateCommand.AddOption(jwtIssuerOption);
+        var jwtRolesOption = new Option<string>("--roles", "A role claim to add to the token. Specify once for each scope. Use a comma to separate scopes.");
+        jwtRolesOption.AddAlias("-r");
+        jwtCreateCommand.AddOption(jwtRolesOption);
+        var jwtScopesOption = new Option<string>("--scopes", "A scope claim to add to the token. Specify once for each scope. Use a comma to separate roles.");
+        jwtScopesOption.AddAlias("-s");
+        jwtCreateCommand.AddOption(jwtScopesOption);
+        var jwtClaimsOption = new Option<string>("--claims", "Claims to add to the token. Specify once for each claim in the format \"name=value\". Use a comma to separate claims.");
```
waldekmastykarz, line 374: Instead of using comma-separated values, let's use the native support for multiple values. That way we don't need parse the string manually and there's less room for error

waldekmastykarz (changes requested):

dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,65 @@ public RootCommand GetRootCommand(ILogger logger)
         var outdatedShortOption = new Option<bool>("--short", "Return version only");
         outdatedCommand.AddOption(outdatedShortOption);
         outdatedCommand.SetHandler(async versionOnly => await OutdatedCommandHandler.CheckVersionAsync(versionOnly, logger), outdatedShortOption);
-
         command.Add(outdatedCommand);
 
+        var jwtCommand = new Command("jwt", "Manage JSON Web Tokens ");
+        var jwtCreateCommand = new Command("create", "Create a new JWT token");
+        var jwtSchemeOption = new Option<string>("--scheme", "The scheme name to use for the token.");
```
waldekmastykarz, line 353: Do we need this option? What else could you put here? How does it influence the generated token?

dev-proxy/CommandHandlers/JwtCommandHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+            ParseClaims(claims)
+        );
+
+        var jwtToken = jwtIssuer.Create(options);
+
+        var jwt = Jwt.Create(
+            scheme,
+            jwtToken,
+            new JwtSecurityTokenHandler().WriteToken(jwtToken),
+            options.Scopes,
```
waldekmastykarz, line 40: Do we want to align with Entra ID which uses the `scp` claim for scopes? Right now, scopes are put in the `scope` claim in the generated token.

dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,65 @@ public RootCommand GetRootCommand(ILogger logger)
...
+        jwtCreateCommand.AddOption(jwtRolesOption);
+        var jwtScopesOption = new Option<IEnumerable<string>>("--scopes", "A scope claim to add to the token. Specify once for each scope. Use a comma to separate roles.")
+        {
+            AllowMultipleArgumentsPerToken = true
+        };
+        jwtScopesOption.AddAlias("-s");
+        jwtCreateCommand.AddOption(jwtScopesOption);
+        var jwtClaimsOption = new Option<IEnumerable<string>>("--claims", "Claims to add to the token. Specify once for each claim in the format \"name=value\". Use a comma to separate claims.")
+        {
+            AllowMultipleArgumentsPerToken = true
```
waldekmastykarz, line 385: Despite the setting, I'm getting an error specifying multiple claims:

image

Seems like the shorthand `-c` collides with the global option for a config file. Let's remove it here.

image

dev-proxy/CommandHandlers/JwtCommandHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+            RandomNumberGenerator.GetBytes(32)
+        );
+
+        var options = new JwtCreatorOptions(
+            scheme,
+            name,
+            audience,
+            issuer,
+            DateTime.UtcNow,
+            DateTime.UtcNow.AddHours(validFor),
```
waldekmastykarz, line 28: Let's update this to minutes 😄

dev-proxy/ProxyHost.cs
```diff
@@ -346,9 +346,65 @@ public RootCommand GetRootCommand(ILogger logger)
...
+        };
+        jwtClaimsOption.AddAlias("-c");
+        jwtCreateCommand.AddOption(jwtClaimsOption);
+        var jwtValidForOption = new Option<double>("--valid-for", "The duration for which the token is valid. Duration is set in minutes.");
+        jwtValidForOption.AddAlias("-v");
+        jwtValidForOption.SetDefaultValue((double)90);
```
waldekmastykarz, line 391: Let's update the default to 60 to align with default Entra ID tokens

dev-proxy/CommandHandlers/JwtCommandHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+            options.Scopes,
+            options.Roles,
+            options.Claims
+        );
+
+        Console.WriteLine(jwt.Token);
+    }
+
+    private static Dictionary<string, string> ParseClaims(IEnumerable<string> claims) 
+        => claims.Select(claim => claim.Split(" ")).ToDictionary(claimParts => claimParts[0], claimParts => claimParts[1]);
```
waldekmastykarz, line 49: Since space collides with the notation for multiple argument values (`--claims value1 value2`), I suggest that we change the notation to `claim:value` which is closer to the value that's in the generated token

waldekmastykarz (changes requested):

waldekmastykarz:
Noticed one more thing: if you specify a claim with the same name as one of the options (eg. `sub`), the token generator adds it to the collection of values rather than replacing the value. In some cases (again `sub`) this leads to incorrect types (`sub` is expected to be a string, but becomes string[]). Not sure if it's an issue, but perhaps something for us to include in remarks for clarity.

Repro:

```
$devproxy jwt create -n Steve -a 123 -r App.Read.All -r App.Write.All -s Foo Bar -i bob --claims "aud 456" --claims "sub Jack"

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1bmlxdWVfbmFtZSI6IlN0ZXZlIiwic3ViIjpbIlN0ZXZlIiwiSmFjayJdLCJqdGkiOiJjNjRiZjFhNCIsInNjb3BlIjpbIkZvbyIsIkJhciJdLCJyb2xlIjpbIkFwcC5SZWFkLkFsbCIsIkFwcC5Xcml0ZS5BbGwiXSwiYXVkIjpbIjQ1NiIsIjEyMyJdLCJuYmYiOjE3MjcwOTc5NjksImV4cCI6MTcyNzQyMTk2OSwiaWF0IjoxNzI3MDk3OTY5LCJpc3MiOiJib2IifQ.7YnArYlXGStA8CQNc0RD_At40wYnfMihsySm4HFcd-c
```

dev-proxy/ProxyHost.cs
```diff
@@ -350,59 +351,79 @@ public RootCommand GetRootCommand(ILogger logger)
...
         jwtAudienceOption.SetDefaultValue("https://myserver.com");
         jwtCreateCommand.AddOption(jwtAudienceOption);
+
         var jwtIssuerOption = new Option<string>("--issuer", "The issuer of the token.");
         jwtIssuerOption.AddAlias("-i");
         jwtIssuerOption.SetDefaultValue("dev-proxy");
         jwtCreateCommand.AddOption(jwtIssuerOption);
-        var jwtRolesOption = new Option<IEnumerable<string>>("--roles", "A role claim to add to the token. Specify once for each scope. Use a comma to separate scopes.")
+
+        var jwtRolesOption = new Option<IEnumerable<string>>("--roles", "A role claim to add to the token. Specify once for each scope.")
```
waldekmastykarz, line 372:
```suggestion
        var jwtRolesOption = new Option<IEnumerable<string>>("--roles", "A role claim to add to the token. Specify once for each role.")
```

waldekmastykarz (changes requested):

waldekmastykarz:
Marking as draft until we add the API

waldekmastykarz (approved):
Awesome!

PR #957, opened 2024-12-12
Add support for generating OpenAPI spec in YAML. Closes #956

### LINKED ISSUE #956: Add support for generating OpenAPI spec in YAML
We currently only support JSON as the format when generating an OpenAPI spec from intercepted requests and responses. We should add support for developers to choose YAML which is a valid format for an OpenAPI spec and is typically the preferred format.

> Basically, JSON does not support comments and requires: commas separating fields, curly brackets around objects, double quotation marks around strings and square brackets around arrays. On the other hand, YAML requires hyphens before array items and relies heavily on indentation, which can be cumbersome on large files (indentation is entirely optional in JSON).
> 
> YAML is typically preferred because of its slightly reduced file size, but the two formats are completely interchangeable (as long as [YAML 1.2](https://en.wikipedia.org/wiki/YAML#Comparison_with_JSON) is used). All examples in these pages will be given in YAML.
> 
> Source: https://learn.openapis.org/specification/structure.html

garrytrinder:
Closes https://github.com/microsoft/dev-proxy/issues/956

### DIFF
--- a/dev-proxy-plugins/RequestLogs/OpenApiSpecGeneratorPlugin.cs
+++ b/dev-proxy-plugins/RequestLogs/OpenApiSpecGeneratorPlugin.cs
@@ -16,6 +16,7 @@ using System.Collections.Specialized;
 using Microsoft.Extensions.Logging;
 using Microsoft.DevProxy.Abstractions.LanguageModel;
 using System.Text.Json.Serialization;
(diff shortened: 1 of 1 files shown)


waldekmastykarz (approved):
Perfect! Nothing to add

PR #958, opened 2024-12-12
Add support for showing timestamps on each intercepted request. Closes #953

### LINKED ISSUE #953: [Feature] Print timestamp for each intercepted request
It would be nice to have the timestamp printed on the console for each intercepted request. This would make it easier to associate requests with log output from our application. Furthermore, if you run multiple tests in a row without restarting DevProxy, it is sometimes hard to identify where the next test run started in the console.

Something like this would be nice:

```
 req   ╭ GET https://contoso.com/some/request/
 time  | 2024-12-05 17:25:12.345
 oops  │ LatencyPlugin: Delaying request for 7000ms
 api   ╰ Passed through
```

garrytrinder:
Closes #953

### DIFF
--- a/dev-proxy/Logging/ProxyConsoleFormatter.cs
+++ b/dev-proxy/Logging/ProxyConsoleFormatter.cs
@@ -48,7 +48,9 @@ public class ProxyConsoleFormatter : ConsoleFormatter
 
         // don't log intercepted response to console
         if (messageType == MessageType.InterceptedResponse ||
-            (messageType == MessageType.Skipped && !_options.ShowSkipMessages))
+            (messageType == MessageType.Skipped && !_options.ShowSkipMessages) ||
+                (messageType == MessageType.Timestamp && !_options.ShowTimestamps)
+            )
         {
             return;
         }
--- a/dev-proxy/Logging/ProxyConsoleFormatter.cs
+++ b/dev-proxy/Logging/ProxyConsoleFormatter.cs
@@ -238,6 +238,7 @@ public class ProxyConsoleFormatter : ConsoleFormatter
             MessageType.Tip => "tip",
             MessageType.Skipped => "skip",
             MessageType.Processed => "proc",
+            MessageType.Timestamp => "time",
             _ => "    "
         };
     }
@@ -258,6 +259,7 @@ public class ProxyConsoleFormatter : ConsoleFormatter
             MessageType.Mocked => (ConsoleColor.DarkMagenta, fgColor),
             MessageType.Failed => (ConsoleColor.DarkRed, fgColor),
             MessageType.Tip => (ConsoleColor.DarkBlue, fgColor),
+            MessageType.Timestamp => (ConsoleColor.Gray, fgColor),
             _ => (bgColor, fgColor)
         };
     }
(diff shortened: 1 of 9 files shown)


dev-proxy/ProxyEngine.cs
```diff
@@ -475,10 +475,17 @@ async Task OnRequestAsync(object sender, SessionEventArgs e)
                 await e.GetRequestBodyAsString();
             }
 
-            using var scope = _logger.BeginScope(e.HttpClient.Request.Method ?? "", e.HttpClient.Request.Url, e.GetHashCode());
+            using var scope = _logger?.BeginScope(e.HttpClient.Request.Method ?? "", e.HttpClient.Request.Url, e.GetHashCode());
```
waldekmastykarz, line 478: I don't think we need to use the null-conditional operator here. We check if _logger isn't null directly in the constructor and it's not-null throughout the code. Let's remove it to simplify the code.

dev-proxy/ProxyEngine.cs
```diff
@@ -475,10 +475,17 @@ async Task OnRequestAsync(object sender, SessionEventArgs e)
...
 
-            using var scope = _logger.BeginScope(e.HttpClient.Request.Method ?? "", e.HttpClient.Request.Url, e.GetHashCode());
+            using var scope = _logger?.BeginScope(e.HttpClient.Request.Method ?? "", e.HttpClient.Request.Url, e.GetHashCode());
 
             e.UserData = e.HttpClient.Request;
-            _logger.LogRequest($"{e.HttpClient.Request.Method} {e.HttpClient.Request.Url}", MessageType.InterceptedRequest, new LoggingContext(e));
+
+            _logger?.LogRequest($"{e.HttpClient.Request.Method} {e.HttpClient.Request.Url}", MessageType.InterceptedRequest, new LoggingContext(e));
+
+            if (_config.ShowTimestamps)
```
waldekmastykarz, line 484: Rather than filtering it here, let's do it in the formatter, like we do for skip-messages:

https://github.com/microsoft/dev-proxy/blob/ea26b9899c4ad001568e9cb4e770d953c8394f09/dev-proxy/Logging/ProxyConsoleFormatter.cs#L50-L54

That way, the code is simpler and the caller doesn't need to be bothered with the conditional logic. This code then simply becomes what's now on line 486.

waldekmastykarz (changes requested):
Let's do a couple of adjustments

dev-proxy/Logging/ProxyConsoleFormatter.cs
```diff
@@ -258,6 +261,7 @@ private static (ConsoleColor bg, ConsoleColor fg) GetMessageTypeColor(MessageTyp
             MessageType.Mocked => (ConsoleColor.DarkMagenta, fgColor),
             MessageType.Failed => (ConsoleColor.DarkRed, fgColor),
             MessageType.Tip => (ConsoleColor.DarkBlue, fgColor),
+            MessageType.Timestamp => (ConsoleColor.Gray, fgColor),
```
waldekmastykarz, line 264: Let's align it with debug, which is `LogLevel.Debug => (bgColor, ConsoleColor.Gray),`, so that it doesn't stand out as much since it's not of primary concern

waldekmastykarz (changes requested):
Let's do one more change 🙂

dev-proxy/Logging/ProxyConsoleFormatter.cs
garrytrinder: image

waldekmastykarz (approved):

PR #1006, opened 2025-02-16
Timeout parameter option

### LINKED ISSUE #845: [Feature Request]: Add Timer/Time Out option to stop dev proxy
It would be nice to add some sort of timer, that devproxy stops itself after that time passes as an additional safety net where it can be left open, in particular in CI/CD scenarios.

```
devproxy --timer 180  
// will stop dev proxy after 3 minutes
```

joshua-siw:
Adds the parameter "--timeout" which when given configures a timer that closes the proxy after it expires.
The Timer resets to the initial value if a request to the proxy is being made.

It uses the same method to stop the Application as the stopProxy Endpoint

dev-proxy-commands-j

Closes #845

Let me know if you have any suggestions :)

### DIFF
--- a/dev-proxy-abstractions/IInactivityTimer.cs
+++ b/dev-proxy-abstractions/IInactivityTimer.cs
@@ -1,0 +1,8 @@
+namespace DevProxy.Abstractions
+{
+    public interface IInactivityTimer
+    {
+        void Reset();
+        void Stop();
+    }
+}
--- a/dev-proxy-abstractions/IProxyConfiguration.cs
+++ b/dev-proxy-abstractions/IProxyConfiguration.cs
@@ -21,4 +21,5 @@ public interface IProxyConfiguration
     IEnumerable<int> WatchPids { get; }
     IEnumerable<string> WatchProcessNames { get; }
     bool ShowTimestamps { get; }
+    long? Timeout { get; }
 }
--- a/dev-proxy/InactivityTimer.cs
+++ b/dev-proxy/InactivityTimer.cs
@@ -1,0 +1,30 @@
+using DevProxy.Abstractions;
+
+namespace DevProxy
+{
+    public class InactivityTimer : IInactivityTimer
+    {
+        private Timer? _timer;
+        private readonly TimeSpan _timeout;
+        private static readonly TimeSpan InfiniteTimeout = TimeSpan.FromMilliseconds(-1);
+
+        public InactivityTimer(long timeoutSeconds, Action timeoutAction)
+        {
+            _timeout = TimeSpan.FromSeconds(timeoutSeconds);
+            Action action = timeoutAction ?? throw new ArgumentNullException(nameof(timeoutAction));
+        
+            _timer = new Timer(_ => action.Invoke(), null, _timeout, InfiniteTimeout);
+        }
+
+        public void Reset()
+        {
+            this._timer?.Change(_timeout, InfiniteTimeout);
+        }
+
+        public void Stop()
+        {
+            this._timer?.Dispose();
+            _timer = null;
+        }
+    }
+}
--- a/dev-proxy/ProxyConfiguration.cs
+++ b/dev-proxy/ProxyConfiguration.cs
@@ -39,5 +39,5 @@ public class ProxyConfiguration : IProxyConfiguration
     public int ApiPort { get; set; } = 8897;
     public bool ShowSkipMessages { get; set; } = true;
     public bool ShowTimestamps { get; set; } = true;
-    public long? Timeout { get; set; } = 60000;
+    public long? Timeout { get; set; }
 }
--- a/dev-proxy/ProxyConfiguration.cs
+++ b/dev-proxy/ProxyConfiguration.cs
@@ -39,4 +39,5 @@ public class ProxyConfiguration : IProxyConfiguration
     public int ApiPort { get; set; } = 8897;
     public bool ShowSkipMessages { get; set; } = true;
     public bool ShowTimestamps { get; set; } = true;
+    public long? Timeout { get; set; } = 60000;
 }
--- a/dev-proxy/ProxyEngine.cs
+++ b/dev-proxy/ProxyEngine.cs
@@ -38,10 +38,7 @@ public class ProxyEngine(IProxyConfiguration config, ISet<UrlToWatch> urlsToWatc
     // Dictionary for plugins to store data between requests
     // the key is HashObject of the SessionEventArgs object
     private readonly Dictionary<int, Dictionary<string, object>> _pluginData = [];
-    private Timer? _inactivityTimeoutTimer;
-    private TimeSpan? InactivityTimeoutTimeSpan => 
-        _config.Timeout.HasValue ? TimeSpan.FromSeconds(_config.Timeout.Value) : null;
-    private static readonly TimeSpan InfiniteTimeout = TimeSpan.FromMilliseconds(-1);
+    private InactivityTimer? _inactivityTimer;
 
     public static X509Certificate2? Certificate => _proxyServer?.CertificateManager.RootCertificate;
 
@@ -82,11 +79,6 @@ public class ProxyEngine(IProxyConfiguration config, ISet<UrlToWatch> urlsToWatc
         var process = new Process() { StartInfo = startInfo };
         process.Start();
         process.WaitForExit();
-    }
-
-    private static void ResetInactivityTimeoutTimer(Timer timer, TimeSpan timeout)
-    {
-        timer.Change(timeout, InfiniteTimeout);
     }
 
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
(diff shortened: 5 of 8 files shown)


dev-proxy/InactivityTimer.cs
```diff
@@ -0,0 +1,30 @@
+using DevProxy.Abstractions;
+
+namespace DevProxy
+{
+    public class InactivityTimer : IInactivityTimer
```
waldekmastykarz, line 5: Let's use the primary constructor here to simplify the code, eg.

```c#
public class InactivityTimer(long timeoutSeconds, Action timeoutAction) : IInactivityTimer
```

dev-proxy-abstractions/IProxyConfiguration.cs
```diff
@@ -21,4 +21,5 @@ public interface IProxyConfiguration
     IEnumerable<int> WatchPids { get; }
     IEnumerable<string> WatchProcessNames { get; }
     bool ShowTimestamps { get; }
+    long? Timeout { get; }
```
waldekmastykarz, line 24: In plugins, such as CachingGuidance or RateLimiting, we append the unit to the config option, to make it clearer to users what to expect, eg. `CacheTresholdSeconds`, `RetryAfterSeconds`, etc. I suggest we do it here too for consistency and clarity, ie. `TimeoutSeconds`. That way, users won't have to wonder what value to put in there.

dev-proxy/InactivityTimer.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        private Timer? _timer;
```
waldekmastykarz, line 7: Since timer is created in the constructor, let's not mark it as nullable. This will help us simplify our code.

dev-proxy/ProxyHost.cs
```diff
@@ -268,6 +270,26 @@ public ProxyHost()
             Arity = ArgumentArity.ZeroOrMore
         };
         _urlsToWatchOption.AddAlias("-u");
+        
+        _timeoutOption = new Option<long?>(TimeoutOptionName, "Time in seconds after which the proxy quits, resets when a request is being made")
```
waldekmastykarz, line 274:
```suggestion
        _timeoutOption = new Option<long?>(TimeoutOptionName, "Time in seconds after which Dev Proxy exits. Resets when Dev Proxy intercepts a request.")
```

dev-proxy/ProxyHost.cs
```diff
@@ -268,6 +270,26 @@ public ProxyHost()
...
+        {
+            ArgumentHelpName = "inactivityTimeout",
+        };
+        _timeoutOption.AddValidator(input =>
+        {
+            try
+            {
+                if (!long.TryParse(input.Tokens[0].Value, out _))
```
waldekmastykarz, line 282: Let's add a check that the value is greater than 0

dev-proxy/ProxyEngine.cs
```diff
@@ -453,6 +460,10 @@ private bool IsProxiedProcess(TunnelConnectSessionEventArgs e)
 
     async Task OnRequestAsync(object sender, SessionEventArgs e)
     {
+        if (_inactivityTimer != null)
```
waldekmastykarz, line 463: Let's simplify this to `_inactivityTimer?.Reset();`

dev-proxy/InactivityTimer.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            _timeout = TimeSpan.FromSeconds(timeoutSeconds);
+            Action action = timeoutAction ?? throw new ArgumentNullException(nameof(timeoutAction));
+        
+            _timer = new Timer(_ => action.Invoke(), null, _timeout, InfiniteTimeout);
+        }
+
+        public void Reset()
+        {
+            this._timer?.Change(_timeout, InfiniteTimeout);
```
waldekmastykarz, line 21: We can remove `this.` here

dev-proxy/InactivityTimer.cs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        public void Stop()
+        {
+            this._timer?.Dispose();
```
waldekmastykarz, line 26: We can remove `this.` here

dev-proxy/InactivityTimer.cs
```diff
@@ -0,0 +1,30 @@
...
+        private Timer? _timer;
+        private readonly TimeSpan _timeout;
+        private static readonly TimeSpan InfiniteTimeout = TimeSpan.FromMilliseconds(-1);
+
+        public InactivityTimer(long timeoutSeconds, Action timeoutAction)
+        {
+            _timeout = TimeSpan.FromSeconds(timeoutSeconds);
+            Action action = timeoutAction ?? throw new ArgumentNullException(nameof(timeoutAction));
+        
+            _timer = new Timer(_ => action.Invoke(), null, _timeout, InfiniteTimeout);
```
waldekmastykarz, line 16: For the InfiniteTimeout, let's use `System.Threading.Timeout.InfiniteTimeSpan` rather than declaring our own property.

dev-proxy/ProxyEngine.cs
```diff
@@ -92,7 +92,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         }
 
         LoadHostNamesFromUrls();
-
+
```
waldekmastykarz, line 95: Seems like an unnecessary change

dev-proxy/ProxyHost.cs
```diff
@@ -268,6 +270,26 @@ public ProxyHost()
             Arity = ArgumentArity.ZeroOrMore
         };
         _urlsToWatchOption.AddAlias("-u");
+        
+        _timeoutOption = new Option<long?>(TimeoutOptionName, "Time in seconds after which the proxy quits, resets when a request is being made")
+        {
+            ArgumentHelpName = "inactivityTimeout",
```
waldekmastykarz, line 276: Let's make it consistent with other args

```suggestion
            ArgumentHelpName = "timeout",
```

dev-proxy/InactivityTimer.cs
```diff
@@ -0,0 +1,30 @@
+using DevProxy.Abstractions;
```
waldekmastykarz, line 1: Let's add the license statement to the top of the file:

```text
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.
// See the LICENSE file in the project root for more information.

```

dev-proxy-abstractions/IInactivityTimer.cs
```diff
@@ -0,0 +1,8 @@
+namespace DevProxy.Abstractions
+{
+    public interface IInactivityTimer
```
waldekmastykarz, line 3: I don't think we need this interface, since we don't really use it anywhere and instantiate InactivityTimer ourself in ProxyEngine.

waldekmastykarz (changes requested):
Very nice! Let's do a couple of small changes before we merge it.

waldekmastykarz:
I marked the PR as draft so that we don't accidentally merge it. Please, mark is as **Ready for review** when you've done the changes @joshua-siw

dev-proxy/ProxyConfiguration.cs
```diff
@@ -39,4 +39,5 @@ public class ProxyConfiguration : IProxyConfiguration
     public int ApiPort { get; set; } = 8897;
     public bool ShowSkipMessages { get; set; } = true;
     public bool ShowTimestamps { get; set; } = true;
+    public long? Timeout { get; set; }
```
waldekmastykarz, line 42: Let's add a reference to this new option in https://github.com/dotnet/dev-proxy/blob/main/schemas/v0.25.0/rc.schema.json so that users will get it in their intellisense.

joshua-siw:
Nice, thanks for the suggestions! Sorry, I am new to GitHub's code management, I’m used to Azure, but will pay attention in the future. I will make the changes at the latest by the weekend.

dev-proxy/InactivityTimer.cs
joshua-siw: yeah my sonarqube was complaining because it couldnt see i was accessing the instance member without this.

waldekmastykarz:
> Nice, thanks for the suggestions! Sorry, I am new to GitHub's code management, I’m used to Azure, but will pay attention in the future. I will make the changes at the latest by the weekend.

All good! We're in this together. Please, don't hesitate to reach out if you have any questions or if we can help with anything.

garrytrinder (changes requested):
Thanks for the updates @joshua-suw, just a minor comment to resolve and we can get this merged! 🚀 Thank you for your contributions, works really well 👏

joshua-siw:
Alright @garrytrinder, i applied and commited your suggestion, thanks for the review😊

garrytrinder (approved):

garrytrinder:
Great work @joshua-siw 👏 Great to have you with us and thank you for your contribution ❤️ This feature will be included in the next release, v0.25.0, which is due this week.

PR #1059, opened 2025-03-12
Renames MinimalPermissionsPlugin to MinimalPermissionsGuidancePlugin. Closes #1058

### LINKED ISSUE #1058: Rename MinimalPermissionsPlugin to MinimalPermissionsGuidancePlugin
To be consistent, we should rename the current MinimalPermissionsPlugin to MinimalPermissionsGuidancePlugin, because it not only checks what permissions are minimal for the set of requests but also compares them against the permissions on the token and reports excessive permissions if any. We could introduce a version that just checks minimal permissions under the current MinimalPermissionsPlugin name.

waldekmastykarz:
Renames MinimalPermissionsPlugin to MinimalPermissionsGuidancePlugin. Closes #1058

### DIFF
--- a/dev-proxy-plugins/RequestLogs/MinimalPermissionsPlugin.cs
+++ b/dev-proxy-plugins/RequestLogs/MinimalPermissionsPlugin.cs
@@ -17,8 +17,6 @@ public class MinimalPermissionsPluginReportApiResult
     public required string[] Requests { get; init; }
     public required string[] TokenPermissions { get; init; }
     public required string[] MinimalPermissions { get; init; }
-    public required string[] ExcessivePermissions { get; init; }
-    public required bool UsesMinimalPermissions { get; init; }
 }
 
 public class MinimalPermissionsPluginReport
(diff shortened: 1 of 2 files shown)


dev-proxy-plugins/RequestLogs/MinimalPermissionsPlugin.cs
garrytrinder, line 1: If this PR is to rename the plugin, why has some of the logic been removed? What's the reason for the change? Is this an expected change?

waldekmastykarz: Originally we released the MinimalPermissionPlugin with the functionality of guidance (check permissions + report excessive). That's inconsistent with other plugins with similar functionality for Graph and APIC. So this PR aligns it with what we have. It renames the old plugin to MinimalPermissionGuidancePlugin, and introduces a new plugin under the old name, with just the permission check functionality. Makes sense?

garrytrinder (approved):

PR #1092, opened 2025-03-31
Updates Docker images to not use a fixed certificate

waldekmastykarz:
To test, build a local image using ./scripts/local-docker.ps1. Then, create a test folder with inside a folder named `cert`.

Then, run:

```sh
docker run --rm -it -p 8000:8000 -p 8897:8897 -v ${PWD}:/config -v ${PWD}/cert:/home/devproxy/.config/dev-proxy/rootCert dev-proxy-local:0.26.0-beta.4
```

After proxy starts, proxy's pfx will appear in the ./cert folder on the host from which it can be trusted.


garrytrinder:
image

PR #1139, opened 2025-04-22
Refactors using language models to support OpenAI-compatible clients

waldekmastykarz:
Refactors using language models to support OpenAI-compatible clients

### DIFF
--- a/dev-proxy-abstractions/LanguageModel/LanguageModelConfiguration.cs
+++ b/dev-proxy-abstractions/LanguageModel/LanguageModelConfiguration.cs
@@ -6,16 +6,16 @@ namespace DevProxy.Abstractions.LanguageModel;
 
 public enum LanguageModelClient
 {
-    LMStudio,
-    Ollama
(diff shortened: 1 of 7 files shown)


dev-proxy-abstractions/LanguageModel/LanguageModelConfiguration.cs
```diff
@@ -6,16 +6,16 @@ namespace DevProxy.Abstractions.LanguageModel;
...
-    LMStudio,
-    Ollama
+    Ollama,
+    OpenAI
 }
 
 public class LanguageModelConfiguration
 {
     public bool CacheResponses { get; set; } = true;
+    public LanguageModelClient Client { get; set; } = LanguageModelClient.OpenAI;
```
waldekmastykarz, line 16: Still uses Ollama by default, but uses its OpenAI-compatible APIs rather than the Ollama-specific ones.

garrytrinder (approved):

Issue #1165, opened 2025-05-06
[BUG]: Add blockUnmockedRequests property of mock response plugin configuration schema


Bartizan:
### Description

In code and documentation, there is the property blockUnmockedRequests that is not included into MockResponsePlugin configuration schema yet.
Nonetheless the blockUnmockedRequests setting can be used and works as documented.

### Expected behaviour

no error in the devproxy console logs

### Actual behaviour

there is a failure in console (but behaviour is not affected) 
>MockResponsePlugin: Plugin configuration validation failed with the following errors: Property 'blockUnmockedRequests' has not been defined and the schema does not allow additional properties. Path 'blockUnmockedRequests', line 4, position 28.

### Steps to reproduce

use blockUnmockedRequests in configuration

### Dev Proxy Version

0.27.0

### Operating system (environment)

Windows

### Shell

PowerShell

### Configuration file

**devproxyrc.json**
```json
...
  "mockResponsePlugin": {
    "$schema": "https://raw.githubusercontent.com/dotnet/dev-proxy/main/schemas/v0.27.0/mockresponseplugin.schema.json",
    "mocksFile": "mocks.json",
    "blockUnmockedRequests": true
  },
...
```

waldekmastykarz:
You're right! Sorry we missed it. We appreciate your fix!

PR #1166, opened 2025-05-06
Adds blockUnmockedRequests property of mock response plugin config schema

Bartizan:
#1165

### DIFF
--- a/schemas/v0.27.0/mockresponseplugin.schema.json
+++ b/schemas/v0.27.0/mockresponseplugin.schema.json
@@ -10,6 +10,10 @@
     "mocksFile": {
       "type": "string",
       "description": "Path to the file containing the mock responses."
+    },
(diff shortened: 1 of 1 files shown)


waldekmastykarz (changes requested):
Could you please add it to all schemas? This feature has been around since v0.12. Thank you!

Bartizan:
The fix is applied to v0.25 and later. 
For v0.24 and earlier there were no "Dev Proxy MockResponsePlugin config schema".
It looks like since v0.25 there was a breaking change and schema mockresponseplugin.schema.json has been renamed to mockresponseplugin.mocksfile.schema.json

waldekmastykarz:
Of course! You're right! For older versions, we only had schemas for the separate files and not config sections. My bad. Thanks for double checking.

waldekmastykarz (approved):

Issue #847, opened 2024-08-13
Remove Dev Proxy CA certificate when uninstalling


garrytrinder:
The installer does not remove the certificate from the store on uninstall.

![image](https://github.com/user-attachments/assets/f587d217-c32e-42e1-b6d6-891173ff5113)

- [ ] Windows
- [ ] macOS

waldekmastykarz:
We should also consider similar clean up on macOS. When you uninstall Proxy using Homebrew, the cert remains in Keychain.

bartizan:
#1208 PR for Windows
Added a certificate remove command and called during uninstallation.

![Image](https://github.com/user-attachments/assets/d4fd9dc0-963e-4dee-97fa-84b19ea409d6)

waldekmastykarz:
Cool! Did you also by any chance look at what we could do (if anything) on macOS with homebrew?

bartizan:
Shouldn't we close the issue or it is under some beta testing now?

PR #1208, opened 2025-05-23
Remove Dev Proxy CA certificate when uninstalling for Win

bartizan:
#847

### DIFF
--- a/dev-proxy/CommandHandlers/CertRemoveCommandHandler.cs
+++ b/dev-proxy/CommandHandlers/CertRemoveCommandHandler.cs
@@ -1,0 +1,26 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace DevProxy.CommandHandlers;
+
+public static class CertRemoveCommandHandler
+{
+    public static void RemoveCert(ILogger logger)
+    {
+        logger.LogTrace("RemoveCert() called");
+
+        try
+        {
+            logger.LogInformation("Uninstalling the root certificate...");
+            ProxyEngine.ProxyServer.CertificateManager.RemoveTrustedRootCertificate(machineTrusted: false);
+            logger.LogInformation("DONE");
+        }
+        catch (Exception ex)
+        {
+            logger.LogError(ex, "Error removing certificate");
+        }
+
+        logger.LogTrace("RemoveCert() finished");
+    }
+}
--- a/dev-proxy/ProxyHost.cs
+++ b/dev-proxy/ProxyHost.cs
@@ -382,7 +382,8 @@ internal class ProxyHost
 
         var sortedCommands = new[]
         {
-            CreateCertEnsureCommand(logger)
+            CreateCertEnsureCommand(logger),
+            CreateCertRemoveCommand(logger),
         }.OrderByName();
 
         certCommand.AddCommands(sortedCommands);
@@ -394,6 +395,13 @@ internal class ProxyHost
         var certEnsureCommand = new Command("ensure", "Ensure certificates are setup (creates root if required). Also makes root certificate trusted.");
         certEnsureCommand.SetHandler(async () => await CertEnsureCommandHandler.EnsureCertAsync(logger));
         return certEnsureCommand;
+    }
+
+    private static Command CreateCertRemoveCommand(ILogger logger)
+    {
+        var certRemoveCommand = new Command("remove", "Remove the certificate from Root Store");
+        certRemoveCommand.SetHandler(() => CertRemoveCommandHandler.RemoveCert(logger));
+        return certRemoveCommand;
     }
 
     private static Command CreateJwtCommand()
--- a/install-beta.iss
+++ b/install-beta.iss
@@ -7,6 +7,7 @@
 #define MyAppVersion "0.28.0-beta.1"
 #define MyAppPublisher ".NET Foundation"
 #define MyAppURL "https://aka.ms/devproxy"
+#define DevProxyExecutable "devproxy.exe"
 
 [Setup]
 ; NOTE: The value of AppId uniquely identifies this application. Do not use the same AppId value in installers for other applications.
@@ -44,6 +45,9 @@ Source: ".\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsu
 
 [UninstallDelete]
 Type:files;Name:"{app}\rootCert.pfx"
+
+[UninstallRun]
+Filename: "{app}\{#DevProxyExecutable}"; Parameters: "cert remove"; RunOnceId: "DelService"; Flags: runhidden; 
 
 [Code]
 procedure RemovePath(Path: string);
--- a/install.iss
+++ b/install.iss
@@ -7,6 +7,7 @@
 #define MyAppVersion "0.28.0"
 #define MyAppPublisher ".NET Foundation"
 #define MyAppURL "https://aka.ms/devproxy"
+#define DevProxyExecutable "devproxy.exe"
 
 [Setup]
 ; NOTE: The value of AppId uniquely identifies this application. Do not use the same AppId value in installers for other applications.
@@ -44,6 +45,9 @@ Source: ".\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsu
 
 [UninstallDelete]
 Type:files;Name:"{app}\rootCert.pfx"
+
+[UninstallRun]
+Filename: "{app}\{#DevProxyExecutable}"; Parameters: "cert remove"; RunOnceId: "DelService"; Flags: runhidden; 
 
 [Code]
 procedure RemovePath(Path: string);


waldekmastykarz:
Cool! We'll check it out asap. Thank you!

waldekmastykarz (changes requested):
Couple of things that I noticed. I'll try to run it asap to see it in action. I like the idea!

dev-proxy/ProxyHost.cs
```diff
@@ -396,6 +397,13 @@ private static Command CreateCertEnsureCommand(ILogger logger)
         return certEnsureCommand;
     }
 
+    private static Command CreateCertRemoveCommand(ILogger logger)
```
waldekmastykarz, line 400: Since this command is removing resources, let's ask for confirmation by default before executing its logic. Users can suppress the prompt by using the `--force` switch.

dev-proxy/CommandHandlers/CertRemoveCommandHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+public static class CertRemoveCommandHandler
+{
+    public static void RemoveCert(ILogger logger)
+    {
+        logger.LogTrace("RemoveCert() called");
+
+        try
+        {
+            logger.LogInformation("Uninstalling the root certificate...");
+            ProxyEngine.ProxyServer.CertificateManager.RemoveTrustedRootCertificate(machineTrusted: false);
```
waldekmastykarz, line 16: I noticed, that even if I say 'No' in the prompt confirming if I want to remove the certificate, it's still getting removed. Let's double check that the code is working as expected.

waldekmastykarz (changes requested):
Works nicely! Let's do a few small fixes before we merge it.

dev-proxy/CommandHandlers/CertRemoveCommandHandler.cs
bartizan: That dialogue is shown by Win API, it appears to confirm removal from Trusted Root CA. A certificate from Personal does not require confirmation.

dev-proxy/ProxyHost.cs
bartizan: To clarify, should we add some confirmation prompt in console and bypass it with `--force` option? 
If we are going to control the appearance of the confirmation dialogue to remove Root certificate... I have to find out if it is feasible.

waldekmastykarz: Yes, when the command is run without `--force`, we should prompt for confirmation, ie. `Do you want to remove the root certificate (y/N)`, `No` being the default. When the command is ran with `--force` we don't show the prompt and proceed with removing the certificate. The prompt should only control if we run the command or exit early. We shouldn't suppress underlying prompts or dialogs.

dev-proxy/CommandHandlers/CertRemoveCommandHandler.cs
waldekmastykarz: If it's a Win API dialog, then let's assume that it's working as intended.

dev-proxy/ProxyHost.cs
bartizan: Implemented. Tested on Windows

install-beta.iss
```diff
@@ -45,6 +46,9 @@ Source: ".\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsu
 [UninstallDelete]
 Type:files;Name:"{app}\rootCert.pfx"
 
+[UninstallRun]
+Filename: "{app}\{#DevProxyExecutable}"; Parameters: "cert remove"; RunOnceId: "RemoveCert"; Flags: runhidden;
```
waldekmastykarz: Since we're uninstalling from the installer and run this command hidden, let's include `--force`. Otherwise, folks won't be able to confirm the prompt

install.iss
```diff
@@ -45,6 +46,9 @@ Source: ".\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsu
 [UninstallDelete]
 Type:files;Name:"{app}\rootCert.pfx"
 
+[UninstallRun]
+Filename: "{app}\{#DevProxyExecutable}"; Parameters: "cert remove"; RunOnceId: "RemoveCert"; Flags: runhidden;
```
waldekmastykarz: Since we're uninstalling from the installer and run this command hidden, let's include `--force`. Otherwise, folks won't be able to confirm the prompt

waldekmastykarz (changes requested):
Very nice and clean implementation for the prompt. Let's add the `--force` switch to the installer invocation so that it executes silently and we'll get it in

install-beta.iss
bartizan: Good catch. 
Fixed

install.iss
bartizan: fixed

bartizan:
@waldekmastykarz
I come to think how a prompt confirmation Y/n should be completed. It could be either by pressing Enter key or just single key (like 'y'or 'n'). It appears this PR implements key + Enter but trust-cert.sh we run to install a certificate uses single key.
In favor of consistency both have to be the same (I see one more fix comming). But which approach to choose? 

The single key is faster, but Key+Enter method sort of safer, gives a better user control and seems to provide adherence to common CLI conventions. So I'd stick with key+enter. What do you think?

PS I am working on MacOS related part of the issue.

waldekmastykarz:
Good point, thank you for bringing it up. Let's do y/n followed by ENTER. I'll create a separate issue for this.

waldekmastykarz:
Hey @bartizan, is this PR ready for review or is there anything else that we need to do?

bartizan:
> Hey @bartizan, is this PR ready for review or is there anything else that we need to do?

it is ready for review. 
Now it should complete the whole issue for Win and Mac.

DevProxy/Proxy/ProxyEngine.cs
```diff
@@ -615,21 +615,44 @@ private static void ToggleSystemProxy(ToggleSystemProxyAction toggle, string? ip
...
+            if (File.Exists(filename))
+            {
+                return false;
+            }
+
+            return CreateFlag();
         }
 
-        try
+        private static bool CreateFlag()
```
waldekmastykarz: Since we already have `Flag` in the class name, let's avoid repeating it here and simply call it `Create`.

DevProxy/Proxy/ProxyEngine.cs
```diff
@@ -615,21 +615,44 @@ private static void ToggleSystemProxy(ToggleSystemProxyAction toggle, string? ip
...
+            catch
+            {
+                return false;
+            }
+            return true;
         }
-        catch { }
 
-        return true;
+        public static void RemoveFlag()
```
waldekmastykarz: Since we already have `Flag` in the class name, let's avoid repeating it here and simply call it `Remove`.

DevProxy/Proxy/ProxyEngine.cs
```diff
@@ -615,21 +615,44 @@ private static void ToggleSystemProxy(ToggleSystemProxyAction toggle, string? ip
         process.WaitForExit();
     }
 
-    private static bool IsFirstRun()
+    internal static class HasRunFlag
```
waldekmastykarz: Let's move this to a separate file for clarity

DevProxy/Commands/CertCommand.cs
```diff
@@ -48,4 +59,85 @@ private async Task EnsureCertAsync()
...
+        {
+            _logger.LogError(ex, "Error removing certificate");
+        }
+        finally
+        {
+            _logger.LogTrace("RemoveCert() finished");
+        }
+    }
+
+    private static bool PromptConfirmation(string message, bool defaultValue)
```
waldekmastykarz: For clarity, let's rename `defaultValue` to `acceptByDefault`

DevProxy/Commands/CertCommand.cs
```diff
@@ -48,4 +59,85 @@ private async Task EnsureCertAsync()
...
+
+    private static void RemoveTrustedCertificateOnMac()
+    {
+        if (!RunTime.IsMac)
+        {
+            return;
+        }
+
+        var bashScriptPath = Path.Join(ProxyUtils.AppFolder, "remove-cert.sh");
+        ProcessStartInfo startInfo = new()
```
waldekmastykarz: For consistency let's change to `var startInfo = new ProcessInfo`

DevProxy/Proxy/ProxyEngine.cs
```diff
@@ -615,21 +615,44 @@ private static void ToggleSystemProxy(ToggleSystemProxyAction toggle, string? ip
...
     }
 
-    private static bool IsFirstRun()
+    internal static class HasRunFlag
     {
-        var firstRunFilePath = Path.Combine(ProxyUtils.AppFolder!, ".hasrun");
-        if (File.Exists(firstRunFilePath))
+        private static readonly string filename = Path.Combine(ProxyUtils.AppFolder!, ".hasrun");
+
+        public static bool IsFirstRun()
```
waldekmastykarz: Since the class is now named `HasRunFlag` let's rename this to `Exists` so that it's cleaner

waldekmastykarz (changes requested):
Nicely done and I like the refactoring of the HasRunFlag to a class with additional functionality. Let's iron out a few wrinkles and we'll get it in. 👏

waldekmastykarz:
Marking as draft. Please mark as "Ready for review" when you've done the changes @bartizan. Thanks!

DevProxy/Proxy/ProxyEngine.cs
bartizan: The Exists is confusing even more.  The func returns boolean depending on wheather the flag is created or not
So I am renaming IsFirstRun() to CreateIfMissing(). Let me know if it is not okay

DevProxy/Proxy/ProxyEngine.cs
bartizan: assumed to move it to Abstractions/Utils/
pls, let me know if there is a more suitable place.

DevProxy/Proxy/ProxyEngine.cs
bartizan: fixed

DevProxy/Proxy/ProxyEngine.cs
bartizan: fixed

DevProxy/Commands/CertCommand.cs
bartizan: fixed

DevProxy/Commands/CertCommand.cs
bartizan: done

DevProxy/Proxy/ProxyEngine.cs
waldekmastykarz: That's also nice. Good idea.

DevProxy/Proxy/ProxyEngine.cs
waldekmastykarz: I moved it back to DevProxy since it's not something that we need to expose outside of the runtime for now.

Issue #1110, opened 2025-04-08
Extend OpenAPISpecGeneratorPlugin to include default values for parameters


garrytrinder:
Azure API endpoints use the `api-version` query string parameter to version endpoints.

When using Dev Proxy to generate an OpenAPI spec the `api-version` parameter in not included in the generated document. This leads to an error when issuing a request as the `api-version` parameter is not sent.

We should look to include this as a parameter in generated documents with a default value.

```yaml
      parameters:
        - name: api-version
          in: query
          required: true
          schema:
            type: string
            default: "2024-11-01-preview"
```

waldekmastykarz:
OK, so I can see the param included in the generated spec but it's missing the default value:

<img width="513" alt="Image" src="https://github.com/user-attachments/assets/1499b1b5-3c49-4a88-9890-e46c4b3690df" />

We can definitely include the specified value as default. What do you think: should we do it by default, or should it be an option, like `currentValueAsDefault`, which is set to `false` by default but when set to `true` copies the values from the intercepted requests as defaults?

I see also that in the generated spec, the param is not marked as required. Do you know if that's necessary for agents to always include the param in the API request?

garrytrinder:
> should we do it by default, or should it be an option, like currentValueAsDefault, which is set to false by default but when set to true copies the values from the intercepted requests as defaults?

I think for ease it would be easier to use the value from the request as the default value. 

Thinking about it some more, should we have an array of known parameters like `api-version` in the configuration, but which developers can override with their own param names?

> I see also that in the generated spec, the param is not marked as required. Do you know if that's necessary for agents to always include the param in the API request?

In this case yes, not sending the param results in an error from the API.

waldekmastykarz:
> I think for ease it would be easier to use the value from the request as the default value.

The reason I'd prefer to expose it via an option and an explicit opt-in is because of potential PII that could end up in the spec.

> Thinking about it some more, should we have an array of known parameters like api-version in the configuration, but which developers can override with their own param names?

We already have a list of standard headers. We could go about it in two ways:

1. Allow customers to specify parameters to exclude and we include everything else
2. Allow customers to specify parameters to include and we exclude everything else

We could offer both options and customers can decide which way they want to go about it.

garrytrinder:
Good shout on PII, then an option makes sense.

I'd say go with an include list, option 2.

bartizan:
working on it. 

Summarizing, We need the following:
[...]

PR #1323, opened 2025-07-11
Extends OpenAPISpecGeneratorPlugin to include default values for parameters

bartizan:
#1110


waldekmastykarz (approved):

DevProxy.Plugins/Generation/OpenApiSpecGeneratorPlugin.cs
```diff
@@ -319,21 +321,35 @@ private void SetParametersFromQueryString(OpenApiOperation operation, NameValueC
         }
 
         Logger.LogDebug("  Processing query string parameters...");
-#pragma warning disable IDE0004
-        var dictionary = (queryParams.AllKeys as string[]).ToDictionary(k => k, k => queryParams[k] as object);
-#pragma warning restore IDE0004
+        var dictionary = queryParams.AllKeys
+            .Where(k => k != null).Cast<string>()
```
waldekmastykarz, line 325:
```suggestion
            .Where(k => k is not null).Cast<string>()
```

Issue #1337, opened 2025-07-21
Rename permissionsToIgnore to permissionsToExclude for GraphMinimalPermissionsGuidancePlugin


bartizan:
Adjust the documentation for permissions configuration property of _GraphMinimalPermissionsGuidancePlugin_. The plugin uses `permissionsToExclude ` property name instead of <strike>permissionsToIgnore</strike>.

Rename permissionsToIgnore to permissionsToExclude.

waldekmastykarz:
Great catch! Thank you for submitting the PR!

PR #1338, opened 2025-07-21
Rename permissionsToIgnore to permissionsToExclude for GraphMinimalPe…

bartizan:
…rmissionsGuidancePlugin schema #1337


waldekmastykarz (changes requested):
Great catch! Let's update this in all schema versions since v0.25.0.

Issue #1342, opened 2025-07-24
Add ability to override default excluded permissions with an empty array in GraphMinimalPermissionsGuidancePlugin


bartizan:
In the GraphMinimalPermissionsGuidancePlugin config, the `permissionsToExclude` property currently defaults to `["profile", "openid", "offline_access", "email"]`. This default applies when the property is undefined, `null`, or an empty array (`[]`).

We need to add the ability to override this default list with an empty list of permissions (meaning none will be excluded).

The suggested behavior is that if the `permissionsToExclude` property is explicitly defined in the configuration, it will completely replace the default permissions. 

Here's a breakdown of the expected behavior for the `permissionsToExclude` property:
| `permissionsToExclude` property | Value | Result |
|--|--|--|
| undefined                                    | absent   | Default list |
| defined                                        | `null`   | `[ ]`  |
| defined                                        | `[ ]`      | `[ ]`  |
| defined                                        | a non-empty, e.g., `["ids.user.read"]`    | `["ids.user.read"]` |

Theoretically, a `null` value for the property could be interpreted as the default permissions, but I'd prefer not to implement that. 

How should we proceed with that?

bartizan:
> Where in the code do you see treating an empty array as the default list? As far as I can see, we only revert to default on undefined/null here:

We add a function to the base plugin class that would resolve a value based on provided property name and its default value. There is we could access IConfiguration interface to check if property exists and its value. 
And add a call of the function somewhere during plugin initialization.

waldekmastykarz:
Let me rephrase, you said: 

> This default applies when the property is undefined, null, or an empty array ([])

But in our code currently I only see the comparison with `null` and not an empty array. Where in our code have you seen us default to `["profile", "openid", "offline_access", "email"]` when the user specifies an empty array?

bartizan:
> ...when the user specifies an empty array?

In brief, it is the line or so you have mentioned above.

The configuration binder works so that if the JSON setting value is null, an empty array ([]), or missing property, the bound value in the settings container is null (actually, it is the initial value of the property which in our case is null... I plan to use an empty list as initial). https://github.com/dotnet/dev-proxy/blob/3d315539082f77df75c5ddb3354bb0a7f5203754/DevProxy.Plugins/Reporting/GraphMinimalPermissionsGuidancePlugin.cs#L36

---
[...]

PR #1347, opened 2025-07-25
1342 allow to set permissions to exclude to empty array overriding defaults

bartizan:
#1342

### DIFF
--- a/DevProxy.Abstractions/Plugins/BasePlugin.cs
+++ b/DevProxy.Abstractions/Plugins/BasePlugin.cs
@@ -99,7 +99,22 @@ public abstract class BasePlugin<TConfiguration>(
     }
     public IConfigurationSection ConfigurationSection { get; } = pluginConfigurationSection;
 
-    protected virtual IEnumerable<string>? GetConfigurationValue(string key, IEnumerable<string>? configuredList, IEnumerable<string>? defaultList = default)
+    /// <summary>
+    /// <para>Evaluates the <paramref name="key"/> array property.
+    ///   If the property exists, the <paramref name="configuredList"/> value is used;
+    ///   otherwise, the default <paramref name="defaultList"/> is applied.</para>
+    /// <para>If the property is <i>null</i>, it is interpreted as an empty array (<i>[]</i>).</para>
+    /// <para>Note: This is necessary because .NET configuration binding cannot differentiate between an empty array,
+    ///   a null value, or a missing property in appsettings.json.
+    ///   See at <see cref="https://github.com/dotnet/runtime/issues/58930"/>
+    /// </para>
+    /// </summary>
+    /// <param name="key">The array property name</param>
+    /// <param name="configuredList">The configured list of string values</param>
+    /// <param name="defaultList">The default list of string values</param>
+    /// <returns>Returns the result list of string values</returns>
+    protected virtual IEnumerable<string>? GetConfigurationValue(string key, IEnumerable<string>? configuredList,
+        IEnumerable<string>? defaultList = default)
     {
         ArgumentNullException.ThrowIfNull(key, nameof(key));
 
--- a/DevProxy.Abstractions/Plugins/BasePlugin.cs
+++ b/DevProxy.Abstractions/Plugins/BasePlugin.cs
@@ -101,7 +101,7 @@ public abstract class BasePlugin<TConfiguration>(
 
     protected virtual IEnumerable<string>? GetConfigurationValue(string key, IEnumerable<string>? list, IEnumerable<string>? defaultList = default)
     {
-        if (key is null) { return null; }
+        ArgumentNullException.ThrowIfNull(key, nameof(key));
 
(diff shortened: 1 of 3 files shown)


DevProxy.Abstractions/Plugins/BasePlugin.cs
```diff
@@ -99,6 +99,15 @@ public TConfiguration Configuration
     }
     public IConfigurationSection ConfigurationSection { get; } = pluginConfigurationSection;
 
+    protected virtual IEnumerable<string>? GetConfigurationValue(string key, IEnumerable<string>? list, IEnumerable<string>? defaultList = default)
+    {
+        if (key is null) { return null; }
```
waldekmastykarz: Key shouldn't be `null`. I suggest we change this to throw an ArgumentNullException.

DevProxy.Abstractions/Plugins/BasePlugin.cs
```diff
@@ -99,6 +99,15 @@ public TConfiguration Configuration
     }
     public IConfigurationSection ConfigurationSection { get; } = pluginConfigurationSection;
 
+    protected virtual IEnumerable<string>? GetConfigurationValue(string key, IEnumerable<string>? list, IEnumerable<string>? defaultList = default)
```
waldekmastykarz: Let's rename `list` to `configuredList` for clarity

DevProxy.Plugins/Reporting/GraphMinimalPermissionsGuidancePlugin.cs
```diff
@@ -33,7 +33,7 @@ public sealed class GraphMinimalPermissionsInfo
 
 public sealed class GraphMinimalPermissionsGuidancePluginConfiguration
 {
-    public IEnumerable<string>? PermissionsToExclude { get; set; }
+    public IEnumerable<string>? PermissionsToExclude { get; set; } = [];
```
waldekmastykarz: Let's not change this, because it changes the default behavior of the plugin.

DevProxy.Abstractions/Plugins/BasePlugin.cs
waldekmastykarz: Let's add a comment why we need this. We can reuse the comment we had in the code previously, in the `GraphMinimalPermissionsGuidancePlugin` class, line 64

waldekmastykarz (changes requested):
Nice! Let's do a few small adjustments before we merge

DevProxy.Abstractions/Plugins/BasePlugin.cs
bartizan: fixed

DevProxy.Abstractions/Plugins/BasePlugin.cs
bartizan: fixed

DevProxy.Plugins/Reporting/GraphMinimalPermissionsGuidancePlugin.cs
bartizan: @waldekmastykarz it's fixed, but I don't get it. 
There is no change from user's perspective. But it is a bit safer approach for development as the list is never null and we don't have to bother to bump into an null-ref error.

DevProxy.Abstractions/Plugins/BasePlugin.cs
bartizan: fixed

DevProxy.Plugins/Reporting/GraphMinimalPermissionsGuidancePlugin.cs
waldekmastykarz: I think I might got confused thinking that if we set `PermissionsToExclude` to `[]` here, we won't exclude anything, but then the property is not set in the config which will cancel this and lead to reverting to default, right? If my thinking is correct, let's align them both and keep them `null`/not set in config, just to make it clearer.

bartizan: sure, let's leave initial value of PermissionsToExclude = null (as it was and as it is currently). 

>  if we set PermissionsToExclude to [] here, we won't exclude anything, but then the property is not set in the config which will cancel this and lead to reverting to default, right?

It is only correct if we change PermissionsToExclude property in _appsettings.json_.
In _GraphMinimalPermissionsGuidancePluginConfiguration_ class, either null or [] initial value does not change the behaviour.

This question bothered me with regards to the following feature where we are adding a similar permissions setting to MinimalPermissionsGuidancePlugin. I will implement it alike there.

waldekmastykarz: > This question bothered me with regards to the following feature where we are adding a similar permissions setting to MinimalPermissionsGuidancePlugin. I will implement it alike there.

Good catch. Let's do it in a separate PR

Issue #1349, opened 2025-07-30
Improves using HttpClient


bartizan:
Improve using of HttpClient.

The preferred and recommended approach is to manage HttpClient lifetimes using IHttpClientFactory via Dependency Injection (DI).

Note: _ApiCenterClient_ is left untouched.

bartizan:
@waldekmastykarz 
Could ApiCenterClient and related handlers be moved to Abstractions project?

waldekmastykarz:
> Could ApiCenterClient and related handlers be moved to Abstractions project?

They shouldn't, because they belong to a specific group of plugins.

PR #1350, opened 2025-07-30
Fixes httpclient usage

bartizan:
#1349


DevProxy.Abstractions/Plugins/BasePlugin.cs
```diff
@@ -75,7 +75,8 @@ public abstract class BasePlugin<TConfiguration>(
     BasePlugin(logger, urlsToWatch), IPlugin<TConfiguration> where TConfiguration : new()
 {
     private TConfiguration? _configuration;
-    private readonly HttpClient _httpClient = httpClient;
+
+    protected HttpClient PluginHttpClient { get; } = httpClient;
```
waldekmastykarz: I'm on the fence if we should do this. On one hand, it's convenient to expose a generic-purpose HttpClient to all plugins since we all have them. On the other hand, I wonder if plugins might need their own HttpClients with specific configuration and it'll get confusing whether they should use the one from the base class or inject their own.

I suggest, we keep the HttpClient in the base class private for now and revisit it in the future if need be.

DevProxy/Program.cs
```diff
@@ -32,12 +32,14 @@ static WebApplication BuildApplication(string[] args, DevProxyConfigOptions opti
 
     return app;
 }
-_ = Announcement.ShowAsync();
```
waldekmastykarz: Originally, we haven't registered Announcement as a service so that we can run it earlier before the whole service container is instrumented. Since it's executed just once in the application, I suggest we keep it that way.

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
```diff
@@ -203,12 +206,18 @@ private static async Task LoadOpenAPIFilesAsync(string folder, ILogger logger, C
...
-                logger.LogError(ex, "Error loading OpenAPI file {FilePath}", filePath);
+                _logger.LogError(ex, "Error loading OpenAPI file {FilePath}", filePath);
             }
         }
     }
+
+    public void Dispose()
+    {
+        _msGraphDbConnection?.Dispose();
+        _httpClient?.Dispose();
```
waldekmastykarz: Since the HttpClient is injected, we shouldn't dispose it ourselves, should we?

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
```diff
@@ -9,17 +9,18 @@
 
 namespace DevProxy.Abstractions.Utils;
 
-public static class MSGraphDbUtils
+public sealed class MSGraphDbUtils(HttpClient httpClient, ILogger<MSGraphDbUtils> logger) : IDisposable
```
waldekmastykarz: Since we're converting this into an actual class, let's rename the class to `MSGraphDb` and drop `Utils`.

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
```diff
@@ -9,17 +9,18 @@
...
+    private readonly Dictionary<string, OpenApiDocument> _openApiDocuments = [];
+    private readonly HttpClient _httpClient = httpClient;
+    private readonly ILogger<MSGraphDbUtils> _logger = logger;
+    private SqliteConnection? _msGraphDbConnection;
 
     // v1 refers to v1 of the db schema, not the graph version
     public static string MSGraphDbFilePath => Path.Combine(ProxyUtils.AppFolder!, "msgraph-openapi-v1.db");
 
-    public static SqliteConnection MSGraphDbConnection
+    public SqliteConnection MSGraphDbConnection
```
waldekmastykarz: Let's rename to `Connection` since it's an instance member now to simplify it

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
```diff
@@ -33,12 +34,12 @@ public static SqliteConnection MSGraphDbConnection
         }
     }
 
-    public static async Task<int> GenerateMSGraphDbAsync(ILogger logger, bool skipIfUpdatedToday, CancellationToken cancellationToken)
+    public async Task<int> GenerateMSGraphDbAsync(bool skipIfUpdatedToday, CancellationToken cancellationToken)
```
waldekmastykarz: Let's rename to `GenerateDbAsync` to simplify it

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
```diff
@@ -48,61 +49,64 @@ public static async Task<int> GenerateMSGraphDbAsync(ILogger logger, bool skipIf
...
 
     private static string GetGraphOpenApiYamlFileName(string version) => $"graph-{version.Replace(".", "_", StringComparison.OrdinalIgnoreCase)}-openapi.yaml";
 
-    private static async Task CreateDbAsync(SqliteConnection dbConnection, ILogger logger, CancellationToken cancellationToken)
+    private async Task CreateDbAsync(CancellationToken cancellationToken)
     {
-        logger.LogInformation("Creating database...");
+        _logger.LogInformation("Creating database...");
+
+        var dbConnection = MSGraphDbConnection;
```
waldekmastykarz: I don't think we need this assignment anymore since we already have it in a class member

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
```diff
@@ -48,61 +49,64 @@ public static async Task<int> GenerateMSGraphDbAsync(ILogger logger, bool skipIf
...
         _ = await createIndex.ExecuteNonQueryAsync(cancellationToken);
     }
 
-    private static async Task FillDataAsync(SqliteConnection dbConnection, ILogger logger, CancellationToken cancellationToken)
+    private async Task FillDataAsync(CancellationToken cancellationToken)
     {
-        logger.LogInformation("Filling database...");
+        _logger.LogInformation("Filling database...");
+
+        var dbConnection = MSGraphDbConnection;
```
waldekmastykarz: I don't think we need this assignment anymore since we already have it in a class member

waldekmastykarz (changes requested):
Nice work! Let's do a couple adjustments before we merge it

waldekmastykarz:
Hey @bartizan, we're releasing a v1 in the next day or two. I'd hate to rush you, but since this PR has some breaking changes, we'd need to get it finalized by the end of the day today or we won't be able to merge it for some time.

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
bartizan: fixed

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
bartizan: fixed

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
bartizan: done

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
bartizan: fixed

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
bartizan: renamed. 
The class is in Utils folder, it suggests itself to move it somewhere else.

DevProxy.Abstractions/Utils/MSGraphDbUtils.cs
bartizan: fixed.

DevProxy/Program.cs
bartizan: reverted

DevProxy.Abstractions/Plugins/BasePlugin.cs
bartizan: reverted

waldekmastykarz (approved):

Issue #1052, opened 2025-03-08
Extend MinimalPermissionsGuidancePlugin with scopes to ignore


waldekmastykarz:
Let's extend the MinimalPermissionsPlugin with the ability to specify permissions to ignore, similarly to how we do it for the [GraphMinimalPermissionsGuidancePlugin](https://learn.microsoft.com/en-us/microsoft-cloud/dev/dev-proxy/technical-reference/graphminimalpermissionsguidanceplugin#configuration-properties)

bartizan:
@waldekmastykarz the issue could be closed (I have no permission on this one).

PR #1365, opened 2025-08-14
Extend MinimalPermissionsGuidancePlugin with scopes to ignore 

bartizan:
#1052


schemas/v0.26.0/minimalpermissionsguidanceplugin.schema.json
```diff
@@ -8,6 +8,14 @@
     },
     "apiSpecsFolderPath": {
       "type": "string"
+    },
+    "permissionsToExclude": {
```
waldekmastykarz: Since we're just adding this property, we can't update previous versions schemas because the property isn't available there. Let's only add it to the latest version

waldekmastykarz (changes requested):
Thanks for the PR. Let's do some changes before we proceed. Since quite a few things changed, I'll have a more thorough look at it in the week of Aug 25. For the future, let's keep PRs focused on the issue at hand and avoid unrelated refactorings. If you see room for improvement, please raise an issue so that we can discuss it first 🙂

DevProxy.Plugins/Utils/GraphUtils.cs
```diff
@@ -96,4 +101,57 @@ internal async Task<IEnumerable<string>> UpdateUserScopesAsync(IEnumerable<strin
 
         return newMinimalScopes;
     }
+
+    public static MethodAndUrl GetMethodAndUrl(string methodAndUrlString)
```
waldekmastykarz: This method doesn't need to be public. Let's make it internal instead

DevProxy.Plugins/Utils/GraphUtils.cs
```diff
@@ -96,4 +101,57 @@ internal async Task<IEnumerable<string>> UpdateUserScopesAsync(IEnumerable<strin
...
+
+        var info = methodAndUrlString.Split(" ");
+        if (info.Length > 2)
+        {
+            info = [info[0], string.Join(" ", info.Skip(1))];
+        }
+        return new(Method: info[0], Url: info[1]);
+    }
+
+    public static string GetTokenizedUrl(string absoluteUrl)
```
waldekmastykarz: This method doesn't need to be public. Let's make it internal instead

DevProxy.Plugins/Utils/GraphUtils.cs
```diff
@@ -96,4 +101,57 @@ internal async Task<IEnumerable<string>> UpdateUserScopesAsync(IEnumerable<strin
...
+    {
+        var sanitizedUrl = ProxyUtils.SanitizeUrl(absoluteUrl);
+        return "/" + string.Concat(new Uri(sanitizedUrl).Segments.Skip(2).Select(Uri.UnescapeDataString));
+    }
+
+    public static MethodAndUrl[] GetRequestsFromBatch(string batchBody, string graphVersion, string graphHostName)
```
waldekmastykarz: This method doesn't need to be public. Let's make it internal instead

DevProxy.Plugins/Utils/GraphUtils.cs
waldekmastykarz: This method isn't related to Graph so let's find a better place for it

bartizan:
> For the future, let's keep PRs focused on the issue at hand and avoid unrelated refactorings. If you see room for improvement, please raise an issue so that we can discuss it first 🙂

At the beginning it seemed like we needed those changes for Graph-related classes. 
But I am okay to cut the unrelated refactorings off the PR to keep it clear.

schemas/v0.26.0/minimalpermissionsguidanceplugin.schema.json
bartizan: fixed

DevProxy.Plugins/Utils/GraphUtils.cs
bartizan: fixed

DevProxy.Plugins/Utils/GraphUtils.cs
bartizan: fixed

DevProxy.Plugins/Utils/GraphUtils.cs
bartizan: fixed

DevProxy.Plugins/Utils/GraphUtils.cs
bartizan: Moved to MethodAndUrlUtils. 
(the method is consumed by graph plugings)

waldekmastykarz:
All good. I'll check it out. It'll just take a slightly bit more time than reviewing the change from the issue 🙂

DevProxy.Plugins/Reporting/MinimalPermissionsGuidancePlugin.cs
```diff
@@ -173,14 +171,45 @@ public override async Task AfterRecordingStopAsync(RecordingArgs e, Cancellation
...
+        {
+            Enabled = false;
+            throw new InvalidOperationException("ApiSpecsFolderPath is required.");
+        }
+
+        Configuration.ApiSpecsFolderPath = ProxyUtils.GetFullPath(Configuration.ApiSpecsFolderPath, ProxyConfiguration.ConfigFile);
+        if (!Path.Exists(Configuration.ApiSpecsFolderPath))
+        {
+            Enabled = false;
+            throw new InvalidOperationException($"ApiSpecsFolderPath '{Configuration.ApiSpecsFolderPath}' does not exist.");
```
waldekmastykarz: This would be better as `FileNotFoundException` (I fixed it in your branch)

PR #1369, opened 2025-08-21
Unobtanium migration

svrooij:
A new proxy backend and a new control flow, each plugin can decide to return a response.

This is a work-in-progress, looking for feedback. Everything changed and not everything is fixed but I think it is the way to go.

Still has several `// TODO: …` notes in the code, but it builds and runs 🤯

### Control flow

The next proxy has two entry points `OnRequest` and `OnResponse`. The `OnRequest` event has to be answered with `ModifyRequest`, `EarlyResponse` or `Continue`. This means that the first plugin that wants to send a response can do do just that.

This also means there is not session or state to modify. The plugin can decide to answer or do nothing. I've split guidance plugins (which will always receive the request since they were not modifying it anyway), from modifying plugins. With the latter if the requests reached that plugin it is not modified before, no need to check if it is modified, just decide what to do.

Is this way workable? This will make the order in which the plugins are added important.

### DevTools plugin

I seem to be doing something wrong, there is no longer any log in the console of the DevTools plugin

### Plugins in general

Plugins can implement the any of the [plugin functions](https://github.com/dotnet/dev-proxy/pull/1369/files#diff-ce63f1fd864e22580c61c045af8e5e39ffffbd7bbe4e398514d55f2ea9594ad7R31-R46), which are all optional, which means the ProxyEngine can [select](https://github.com/dotnet/dev-proxy/pull/1369/files#diff-580e0d2e7da72a48b221bc8804906ace000ef2e1234d3c09fa5492fb7b8d5176R419) just the plugins that implement that function.

### Plugin URLS

Maybe plugins should optionally have a list of urls for that plugin. Then we could skip plugins that don't do stuff with the current url.


DevProxy.Abstractions/Plugins/BasePlugin.cs
```diff
@@ -17,49 +17,57 @@ public abstract class BasePlugin(
...
+    /// <inheritdoc/>
     public abstract string Name { get; }
+    protected ILogger Logger { get; } = logger;
+    /// <inheritdoc/>
+    public virtual Func<RequestArguments, CancellationToken, Task<PluginResponse>>? OnRequestAsync { get; }
 
-    public virtual Option[] GetOptions() => [];
-    public virtual Command[] GetCommands() => [];
+    /// <inheritdoc/>
+    public virtual Func<RequestArguments, CancellationToken, Task>? ProvideRequestGuidanceAsync { get; }
```
waldekmastykarz, line 34: What's the benefit of using delegates vs. virtual methods that we can override? Since we need to implement the logic anyway, isn't using a delegate an unnecessary complexity? Ie. I need to implement the method and add a reference to it in the delegate.

DevProxy.Abstractions/Utils/ProxyUtils.cs
```diff
@@ -158,7 +157,7 @@ public static IList<MockResponseHeader> BuildGraphResponseHeaders(Request reques
                 new ("Date", requestDate),
                 new ("Content-Type", "application/json")
             };
-        if (request.Headers.FirstOrDefault((h) => h.Name.Equals("Origin", StringComparison.OrdinalIgnoreCase)) is not null)
+        if (request.Headers.Contains("Origin"))
```
waldekmastykarz, line 160: Is this comparison case-insensitive?

DevProxy.Abstractions/Plugins/BasePlugin.cs
svrooij: The benefit is performance, before calling the plugin we already know if it implemented this. So in the ProxyEngine it skips plugins that did not implement this. We can do the null check on the function, instead of the base plugin returning completed task

And on the other hand, this way we force the plugin to think about what it is doing next, early response/modify request/continue as normal.

DevProxy.Abstractions/Utils/ProxyUtils.cs
svrooij: ~~According to the http specs the headers should be case insensitive, but they come through as they come through. Kestrel parses them to an actual object no matter the casing. So I thought let's make it case insensitive here as well.~~

No you're right that might be a mistake, let me check

DevProxy.Plugins/Reporting/UrlDiscoveryPlugin.cs
```diff
@@ -28,19 +29,18 @@ public override Task AfterRecordingStopAsync(RecordingArgs e, CancellationToken
         }
 
         var requestLogs = e.RequestLogs
-            .Where(l => ProxyUtils.MatchesUrlToWatch(UrlsToWatch, l.Context?.Session.HttpClient.Request.RequestUri.AbsoluteUri ?? ""));
+            .Where(l => ProxyUtils.MatchesUrlToWatch(UrlsToWatch, l.Request!.RequestUri!.AbsoluteUri ?? ""));
```
waldekmastykarz, line 32: The `!.` looks a bit messy. If it can be null, we need to do more robust checks to avoid runtime failures.

waldekmastykarz:
We're going to close this one in favor of the migration to Kestrel we want to give a shot. Thanks again for this suggestion and working with us all this time, shaping Dev Proxy to what it is today.

Issue #1023, opened 2025-03-01
Extend MinimalPermissionsPlugin to support delegated and application permissions


waldekmastykarz:
Extend MinimalPermissionsPlugin to support delegated and application permissions. To do this, we need to introduce a new property named `schemeName`. If specified, the plugin retrieves scopes from that scheme. If not specified, plugin reads information from the first oauth2 scheme in the spec, like it does now.

bartizan:
> `schema` and `scheme` are two different things so I suggest we stick to `scheme`. If `schemeName` is too confusing, then we can use `scheme` to better align with OpenAPI docs. Thoughts?

I tend to confuse 'scheme' and 'schema' but I agree that keeping it as schemeName seems to be the best compromise.

I considered renaming the parameter to a more definitive name, like 'schemeKey', 'schemeDefininitionName' or 'schemeDefinition', to point at the security scheme object in the spec instead of the scheme name field. The scheme object has a "name" field that is applied for an apiKey scheme, which is the name of the header, query or cookie parameter to be used.

PR #1391, opened 2025-09-19
Extend MinimalPermissionsPlugin by introducing the schemeName property

bartizan:
#1023


DevProxy.Plugins/Extensions/OpenApiDocumentExtensions.cs
```diff
@@ -16,6 +16,11 @@ namespace Microsoft.OpenApi.Models;
 static class OpenApiDocumentExtensions
 {
     public static ApiPermissionsInfo CheckMinimalPermissions(this OpenApiDocument openApiDocument, IEnumerable<RequestLog> requests, ILogger logger)
+    {
+        return CheckMinimalPermissions(openApiDocument, requests, logger, null);
+    }
+
+    public static ApiPermissionsInfo CheckMinimalPermissions(this OpenApiDocument openApiDocument, IEnumerable<RequestLog> requests, ILogger logger, string? schemeName)
```
waldekmastykarz: Since we're adding an optional parameter, couldn't we add it to the original signature rather than introducing a new method, similarly to what we've done with `GetEffectiveScopes`?

DevProxy.Plugins/Extensions/OpenApiDocumentExtensions.cs
```diff
@@ -210,11 +222,17 @@ public static string[] GetEffectiveScopes(this OpenApiOperation operation, OpenA
...
-    public static OpenApiSecurityScheme[] GetOAuth2Schemes(this OpenApiDocument openApiDocument)
+    public static OpenApiSecurityScheme[] GetOAuth2Schemes(this OpenApiDocument openApiDocument, string? schemeName)
     {
-        return [.. openApiDocument.Components.SecuritySchemes
-            .Where(s => s.Value.Type == SecuritySchemeType.OAuth2)
-            .Select(s => s.Value)];
+        var schemes = openApiDocument.Components.SecuritySchemes
+            .Where(s => s.Value.Type == SecuritySchemeType.OAuth2);
+
+        if (!string.IsNullOrWhiteSpace(schemeName))
```
waldekmastykarz: Why not add the name check to the `.Where` query above?

DevProxy.Plugins/Reporting/MinimalPermissionsPlugin.cs
```diff
@@ -132,14 +134,22 @@ public override async Task AfterRecordingStopAsync(RecordingArgs e, Cancellation
...
+                    Configuration.SchemeName, string.Join(", ", result.MinimalPermissions));
+            }
         }
 
         var report = new MinimalPermissionsPluginReport()
         {
             Results = [.. results],
             UnmatchedRequests = [.. unmatchedRequests],
-            Errors = [.. errors]
+            Errors = [.. errors],
```
waldekmastykarz: Let's remove this dangling comma

DevProxy.Plugins/Reporting/MinimalPermissionsPluginReport.cs
```diff
@@ -36,9 +37,18 @@ public sealed class MinimalPermissionsPluginReport : IMarkdownReport, IPlainText
                 .AppendLine("### Requests")
                 .AppendLine()
                 .AppendJoin(Environment.NewLine, apiResult.Requests.Select(r => $"- {r}"))
-                .AppendLine()
+                .AppendLine();
 
-                .AppendLine()
+            if (!string.IsNullOrWhiteSpace(apiResult.SchemeName))
```
waldekmastykarz: Rather than introducing a whole separate section for just one name, let's add it to the `### Minimal permissions` heading, like `### Minimal permissions for xyz scheme`

DevProxy.Plugins/Extensions/OpenApiDocumentExtensions.cs
bartizan: fixed

DevProxy.Plugins/Reporting/MinimalPermissionsPlugin.cs
bartizan: fixed

DevProxy.Plugins/Extensions/OpenApiDocumentExtensions.cs
bartizan: Left unchanged.
If we look at the parameter a bit closely, it's not an optional. There are 2 separate methods as it helps to avoid an ambiguity during refactoring. No change comming here as the extra method is going away in a few days.

DevProxy.Plugins/Reporting/MinimalPermissionsPluginReport.cs
bartizan: fixed

DevProxy.Plugins/Extensions/OpenApiDocumentExtensions.cs
waldekmastykarz: > If we look at the parameter a bit closely, it's not an optional.

I don't quite understand it. `schemeName` is defined as `string?` so when not specified, it will default to `null`,  and you can invoke the method without it, no?

bartizan: I presume here is a mix-up with syntax from other language, e.g., TypeScript:
A TS function: 
```typescript
foo(name: string, schemeName?: string); 
//schemeName - is the optional param (type: string | undefined)
```

in C# optional parameters are defined by assigning a default value directly in the parameter declaration.
A C# function: 
```csharp
foo(string? name,                   // Type: nullable string (string?)
    string? schemeName = default);  // Type: nullable string (string?); Default: null
```
sample of a call: `foo("John")`, `foo("John", "oidc")`, or `foo("John", null)`

waldekmastykarz: Duh! You're totally right, obviously! Wouldn't it make sense though to combine both methods and change the signature to `string? scheme = null`?

bartizan: fixed

Issue #1414, opened 2025-10-07
Extend OpenAITelemetryPluginReport with cached tokens information


waldekmastykarz:
Extend OpenAITelemetryPluginReport with cached tokens information. The information is available in the `response.body.usage.prompt_tokens_details.cached_tokens` property.

PR #1439, opened 2025-11-11
Extend OpenAITelemetryPluginReport with cached tokens information

bartizan:
#1414


waldekmastykarz (approved):
Brilliant! Nothing to add

Issue #1440, opened 2025-11-11
Make OpenAITelemetryPlugin record report metrics only if 'recording' is on


bartizan:
The OpenAITelemetryPlugin should only record some report metrics when the 'recording' state is active.
Currently, all records are collected and compiled into a report after recoding is stopped.

waldekmastykarz:
We chose to send metrics without recording because we're sending them live, as we're intercepting requests. Is there a reason why you'd want to capture metrics only while recording?

waldekmastykarz:
Ah, understood. Yeah, that makes sense. Rather than recording the usage for the report as we're intercepting requests, we could do it only based on the requests we recorded. You're right. Thanks for pointing it out.

PR #1450, opened 2025-11-17
Make OpenAITelemetryPlugin record report metrics only if 'recording' is on

bartizan:
#1440


waldekmastykarz (changes requested):
Rather than introducing a whole new way of handling recording, let's use the existing approach that we have, for example here: https://github.com/dotnet/dev-proxy/blob/5fc6944fac9421d54a1bd8fe932fb52717320419/DevProxy.Plugins/Generation/HarGeneratorPlugin.cs#L38. It will mean some more changes in the plugin but it'll be consistent with other plugins and fewer changes to the core functionality.

PR #1481, opened 2025-12-26
Expose apiPort via commndline

### LINKED ISSUE #1472: Expose the apiPort setting as a command line option
Let's expose the `apiPort` setting as a command line option, like we do with the `port`. It'll allow customers to dynamically configure Dev Proxy in case they need a unique port for the API that doesn't collide with other parts of the solution.

- [ ] Update code
- [ ] Update docs @ https://learn.microsoft.com/en-us/microsoft-cloud/dev/dev-proxy/technical-reference/proxy-settings

niteshsinghal85:
Fixes #1472 

Added support for configuring the Dev Proxy API port via the --api-port command-line option. 
This allows users to dynamically set the API port at runtime instead of only through configuration files. 

The change makes IProxyConfiguration.ApiPort settable and integrates the new option into the command-line parsing logic with a default value of 8897.

### DIFF
--- a/DevProxy.Abstractions/Plugins/BaseLoader.cs
+++ b/DevProxy.Abstractions/Plugins/BaseLoader.cs
@@ -79,7 +79,9 @@ public abstract class BaseLoader(HttpClient httpClient, ILogger logger, IProxyCo
         using var document = JsonDocument.Parse(fileContents, ProxyUtils.JsonDocumentOptions);
         var root = document.RootElement;
 
-        if (!root.TryGetProperty("$schema", out var schemaUrlElement))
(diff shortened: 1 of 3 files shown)


waldekmastykarz:
Cool! We'll review in the next couple of days. Thanks!

niteshsinghal85:
Hi @waldekmastykarz @garrytrinder 

Where and how I do sign this, 

Contributor License Agreement is not agreed yet.

niteshsinghal85:
@dotnet-policy-service agree

garrytrinder:
Yes, now I can, I guess you found the link 😊

PR #1652, opened 2026-05-12
Add metadata section with version number

garrytrinder:


garrytrinder:
**Why metadata?**

`version` is not a supported property in VS Code

image

PR #1742, opened 2026-07-03
Update dependent packages lock files

waldekmastykarz:
The recent NuGet bumps (Markdig, Scriban) in DevProxy.Abstractions updated its lock file but not the dependent lock files in DevProxy and DevProxy.Plugins, causing CI to fail with NU1004 errors on main and all open PRs.

This regenerates the lock files for DevProxy and DevProxy.Plugins to fix the issue.


garrytrinder (approved):

waldekmastykarz:
Closing: no modified files (lock files are already up to date).

PR #1877, opened 2026-09-21
Support watching child process traffic

### LINKED ISSUE #1800: Feature request: Intercept traffic from a process + all children processes
Thank you for this great project. This is a feature request to expand capabilities documented in https://learn.microsoft.com/en-us/microsoft-cloud/dev/dev-proxy/how-to/intercept-requests-from-specific-processes, to capture all children processes for a given process.

Typical use case: Capture only the traffic from Microsoft Scout, currently not easy because Scout spawns many child processes (screenshot below shows some, but there are more depending on scenarios):
Image

waldekmastykarz:
## Summary

- add opt-in `--watch-process-tree` / `watchProcessTree` support
- match connection owners against configured PID or process-name ancestors
- cache decisions by PID and process start time to avoid PID-reuse errors
- resolve process ancestry on Windows and Unix-like systems
- update the v4 schema, configuration reference, and tests

Closes #1800

## Validation

- `dotnet build DevProxy/DevProxy.csproj`
- `dotnet test DevProxy.Proxy.Kestrel.Tests/DevProxy.Proxy.Kestrel.Tests.csproj --no-restore`
- `dotnet test DevProxy.Tests/DevProxy.Tests.csproj`
- detached proxy smoke test with `--watch-process-tree`


DevProxy.Proxy.Kestrel/Internal/ProcessFilter.cs
```diff
@@ -75,30 +96,280 @@ public bool IsWatchedProcess(int clientPort)
...
+            {
+                if (TryGetCachedDecision(parentKey.Value, out var cached))
+                {
+                    isWatched = cached;
+                    break;
+                }
+                cacheKeys.Add(parentKey.Value);
+            }
+
+            if (_pids.Contains(parentPid) || MatchesProcessName(parentPid))
```
garrytrinder, line 155: On Windows, `PROCESSENTRY32.th32ParentProcessID` is the creator PID and can refer to an unrelated process after the original parent exits and that PID is reused. This walk resolves metadata for whichever process currently owns `parentPid`, then matches its configured PID/name, so traffic from an unrelated process can be classified as a descendant and decrypted. Please validate each edge using process start times (the parent must predate the child) and add a test where an ancestor PID has been reused.

DevProxy.Proxy.Kestrel/Internal/ProcessFilter.cs
```diff
@@ -75,30 +96,280 @@ public bool IsWatchedProcess(int clientPort)
...
+            // Process exited or its metadata is unavailable.
             return null;
         }
     }
+
+    private static DateTimeOffset? DefaultResolveStartTime(int pid)
+    {
+        try
+        {
+            return Process.GetProcessById(pid).StartTime.ToUniversalTime();
```
garrytrinder: `Process.GetProcessById` returns an `IDisposable` `Process`, and accessing `StartTime`/`ProcessName` can open OS handles. This new resolver runs for the connection owner and each ancestor on uncached decisions, leaving those handles for finalization. Please use `using var process = Process.GetProcessById(pid);` here and in `DefaultResolveName`.

garrytrinder (changes requested):
Requesting changes for the Windows ancestor PID-reuse issue. The process-handle disposal issue should also be addressed while touching this path.

DevProxy.Proxy.Kestrel/Internal/ProcessFilter.cs
waldekmastykarz: Thanks, fixed. Both `DefaultResolveName` and `DefaultResolveStartTime` now dispose the `Process` instance immediately with `using var` after reading the metadata.

DevProxy.Proxy.Kestrel/Internal/ProcessFilter.cs
waldekmastykarz: Good catch, thanks. I updated the walk to validate every parent edge using the process start times before matching or using a cached decision. If the current owner of the parent PID started after the child, or either start time can't be resolved, the walk stops and the process isn't treated as a descendant. I allow equal timestamps because process start-time granularity varies by platform. I also added coverage for a reused ancestor PID.

DevProxy.Abstractions/Proxy/IProxyConfiguration.cs
```diff
@@ -49,4 +49,5 @@ public interface IProxyConfiguration
     bool ValidateSchemas { get; }
     IEnumerable<int> WatchPids { get; set; }
     IEnumerable<string> WatchProcessNames { get; set; }
+    bool WatchProcessTree { get; set; }
```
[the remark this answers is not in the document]
waldekmastykarz: Good point, thanks. I moved `WatchProcessTree` to a separate optional `IProcessTreeProxyConfiguration` interface instead of adding it to `IProxyConfiguration`. Existing external implementations therefore continue to compile unchanged, while the built-in configuration opts into the capability explicitly. I preferred this over a default no-op setter so an attempted opt-in can't be silently ignored.

garrytrinder (approved):

